PR #28, opened 2023-09-06
Set executable flag before tar command

inf17101:
Set the executable flag with chmod before binaries are placed into the tar.gz.

# Definition of Done

- [ ] conform to [unit verification strategy](https://eclipse-ankaios.github.io/ankaios/development/unit-verification/)

### DIFF
--- a/tools/create_artifacts.sh
+++ b/tools/create_artifacts.sh
@@ -37,6 +37,7 @@ ANK_BIN_DIR="${DIST_DIR}/bin"
 RELEASE_FILE_NAME_BASE="ankaios-${RELEASE_ARCHITECTURE}"
 
 echo "Creating the archive for '$RELEASE_FILE_NAME_BASE'"
+chmod +x "${ANK_BIN_DIR}"/ank{,-server,-agent}
 cd "${DIST_DIR}"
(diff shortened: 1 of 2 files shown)


tools/create_artifacts.sh
```diff
@@ -37,6 +37,7 @@ ANK_BIN_DIR="${DIST_DIR}/bin"
 RELEASE_FILE_NAME_BASE="ankaios-${RELEASE_ARCHITECTURE}"
 
 echo "Creating the archive for '$RELEASE_FILE_NAME_BASE'"
+chmod +x "${ANK_BIN_DIR}"/ank{,-server,-agent}
```
windsource, line 40: The contents of this directory will finally be copied to /usr/local/bin so all files shall be executable. Why not just

```
chmod +x "${ANK_BIN_DIR}/*"
```

inf17101: Due to security reasons we shall not chmod blind everything that is inside a folder with a wildcard expression. This was also discussed in the meeting yesterday.

windsource: OK for me.

windsource (approved):
LGTM 👍

inf17101:
Fork was executed successfully:
image

Also the binaries are now executable when downloading and extracting the tar.gz

inf17101:
Please do not merge the PR now because then the install script is changed and does not set the executable flag anymore.

Issue #9, opened 2023-09-05
Unify and improve error propagation and handling


windsource:
Reported by @krucod3:

## Description

We have started with proper error propagation and handling in some places (e.g. try_from for proto objects), but there are still some places where this was not done (e.g. `impl From<i32> for ExecutionState `). 
We should go through the code and see if:
* the `Err` we return are correct
* conversions do not call unwraps 
* default mappings for conversions are reasonable
* The error handling is consistent (the 4 proxy command all return different errors. Especially `state_change_proxy::forward_from_proto_to_ankaios` should be changed to be close to the error handling in `execution_command_proxy`)

## Goals

* Update the error handling in Ankaios s.t. a common concept is used throughout the code
* Update the Rust coding guidelines

## Tasks

tbd.

inf17101:
Inside unit tests we keep the unwrap calls, reason: see [coding guidelines](https://eclipse-ankaios.github.io/ankaios/development/rust-coding-guidelines/#avoid-unwrap-and-expect).

After discussion with @christoph-hamm:
For now, we keep also the unwrap calls for acquiring the lock of a mutex, because the rust doc says that the panic happens only if another thread holding the mutex does a panic.

I am currently thinking about how to fix the unwrap calls for the async awaits and for channel's send/recv calls. Of course we can put a conditional check, but this implies introducing more nested checks with decreasing the readability.

inf17101:
For `From<i32> for ExecutionState` for me it is not clear why we shall introduce an Result here for error propagation. The code shows that every input number (i32) is mapped to a value, so there is no code path that is taken in case of an error.
On the other hand `std::str::FromStr for ExecutionState` does have return type Result, but also does not have a code path that is returning an error. Results/Options shall not be used if we do not return an error, it is just syntactic sugar then... But maybe I misunderstood something here. My suggestion is to return the unwrapped value.

inf17101:
For `impl From<i32> for ExecutionState` I would not introduce a `Result<T,E>`return type. Every code path returns a valid value. Changing the result type would add unnecessary complexity when using/calling the conversion. Any objections on this?

inf17101:
Like discussed with the team in our Rust dojo session about error handling, a generic introduction of Error Type boundaries shall be done later , not with this ticket (cost/time balance). A separate issue was opened by me: https://github.com/eclipse-ankaios/ankaios/issues/46

krucod3:
Work here is done.

PR #30, opened 2023-09-07
Improve config file handling

flba-eb:
- In case the startup configuration file cannot be read/parsed, print out an error and return without `panic`ing as this is a handled error
- Write more clear error message for most common failure: file not found (and wrong file type)
- Write message to `log`
- Slightly rephrase description of "--startup-config" option

# Definition of Done

- [x] requirements are up-to-date and requirements for new features have been added and mapped to source and test
- [x] conform to [unit verification strategy](https://eclipse-ankaios.github.io/ankaios/development/unit-verification/)

### DIFF
--- a/server/src/main.rs
+++ b/server/src/main.rs
@@ -37,12 +37,24 @@ async fn main() -> Result<(), BoxedStdError> {
 
     let args = cli::parse();
 
-    let data = fs::read_to_string(args.path)
-        .unwrap_or_else(|error| panic!("Could not read the startup state.\nError: {error}"));
+    let data = fs::read_to_string(args.path).unwrap_or_else(|error| {
+        use std::io::ErrorKind as EK;
+        match error.kind() {
+            EK::NotFound => log::error!("Startup configuration file not found."),
+            EK::InvalidData => log::error!("Startup configuration file contains non UTF-8 data."),
+            e => log::error!("Error reading startup configuration file: {}.", e),
+        }
+        std::process::exit(1);
+    });
     // [impl->swdd~server-state-in-memory~1]
     // [impl->swdd~server-loads-startup-state-file~1]
-    let state: State = state_parser::parse(data)
-        .unwrap_or_else(|error| panic!("Parsing start config failed with error {}", error));
+    let state: State = state_parser::parse(data).unwrap_or_else(|error| {
+        log::error!(
+            "Parsing start configuration file failed with error: {}.",
+            error
+        );
+        std::process::exit(1);
+    });
     log::debug!(
         "The state is initialized with the following workloads: {:?}",
         state.workloads
--- a/server/src/cli.rs
+++ b/server/src/cli.rs
@@ -26,7 +26,7 @@ pub fn parse() -> Arguments {
         about="Ankaios - your friendly automotive workload orchestrator.\nWhat can the server do for you?")]
 pub struct Arguments {
     #[clap(short = 'c', long = "startup-config")]
-    /// The path to the startup config yaml.
+    /// The path to the startup configuration file in yaml format.
     pub path: String,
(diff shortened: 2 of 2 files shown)


server/src/main.rs
```diff
@@ -37,12 +37,24 @@ async fn main() -> Result<(), BoxedStdError> {
...
     let args = cli::parse();
 
-    let data = fs::read_to_string(args.path)
-        .unwrap_or_else(|error| panic!("Could not read the startup state.\nError: {error}"));
+    let data = fs::read_to_string(args.path).unwrap_or_else(|error| {
+        use std::io::ErrorKind as EK;
+        match error.kind() {
+            EK::NotFound => log::error!("Startup configuration file not found."),
+            EK::InvalidData => log::error!("Startup configuration file contains non UTF-8 data."),
+            e => log::error!("Error reading startup configuration file: {}.", e),
```
krucod3, line 45: Outputting the error as in the last case would actually already do the job without overcomplicating things.
Additionally it would allow to unify the handling and avoiding repetition.

server/src/main.rs
```diff
@@ -37,12 +37,24 @@ async fn main() -> Result<(), BoxedStdError> {
 
     let args = cli::parse();
 
-    let data = fs::read_to_string(args.path)
-        .unwrap_or_else(|error| panic!("Could not read the startup state.\nError: {error}"));
+    let data = fs::read_to_string(args.path).unwrap_or_else(|error| {
+        use std::io::ErrorKind as EK;
```
krucod3, line 41: I don't really think this helps with the readability.

krucod3 (commented):
@flba-eb, first of all thank you very much for opening the PR and for your efforts to improve Ankaios!
You are absolutely right that the current handling of panicking is not correct and should be exchanged with a graceful exit. 

There are two other tickets related to the topic: #9 and #34
Especially #9 is about the general error handling and after seeing your ticket we started thinking about adding common functionality to handle these exit conditions. 

The current idea is to add a trait that extends Result and adds an expect_or_exit function that logs and exits. This would also solve the reuse topic mentioned in the comments.

Would you be interested in helping us with this common functionality? If yes, I would be happy to discuss some details with you. If not, we can also add it with the efforts for #9. 

P.S.:
I see now that we don't have a checkbox for compliance with the development guidelines and especially the [coding guidelines](https://eclipse-ankaios.github.io/ankaios/development/rust-coding-guidelines/). We will add an issue for updating the definition of done, but for now you can have a peek in there if you are interested.

server/src/main.rs
flba-eb: Even as a programmer I have to search for the error cause in this message, there is a lot of noise included:
```text
thread 'main' panicked at 'Parsing start config failed with error missing field `workloads`', server/src/main.rs:52:33
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```
Using logging output also has some noise at the beginning (timestamp), but at least to me its easier to read:
```text
[2023-09-07T15:29:58Z ERROR ank_server] Parsing start configuration file failed with error: missing field `workloads`.
```
Originally I `eprintln`ed the text out (which would be consistent with most other applications) but decided to be consistent with the remaining ankaios code.

flba-eb:
Can be closed in favor of https://github.com/eclipse-ankaios/ankaios/issues/9.

server/src/main.rs
krucod3: Sorry for the confusion, I meant the aliasing of ErrorKind to EK.

krucod3:
As agreed, the proposed improvements here will be done with the changes for #9 in a more generic way that can be reused throughout the entire codebase.

@flba-eb, thank you very much for opening the PR and the ideas!

Issue #24, opened 2023-09-05
Allow `command` entry in Podman RuntimeConfig to overrides image's entrypoint


windsource:
Reported by @christoph-hamm:

## Description

The `command` entry in the Podman RuntimeConfig shall override the `entrypoint` of the Podman image. At the moment `command` entry is combined with `args` entry and used as `command` argument when creating the Podman container.

The naming here is little bit confusing as `command` means something different in the RuntimeConfig, than in the Podman API. The terms `command` and `args`, as used in the RuntimeConfig, have been taken from the Kubernetes Containers definition.

## Goals

It is possible to override the `Entrypoint` of a Podman image, using the `command` entry in the RuntimeConfig.

## Tasks

- [x] Move Podman-related functions (i.e. functions using Podman API) to the `podman_utils.rs`. PR:  #40 
- [x] Forward the value of `command` entry of the RuntimeConfig to the `entrypoint` parameter when creating a Podman container - #45 
- [x] Keep forward the value of `args` entry of the RuntimeConfig to the `command` parameter when creating a Podman container - #45 
- [x] Fix the unit tests - unit tests will be created in #1 
- [x] Checks if it is now possible to override the Podman image's `Entrypoint` - see info below

maturar:
This ticket is a chance for small refactoring. We would like to move all functions which call Podman API from the `workload.rs` to the `podman_utils.rs`. The motivation is to have podman-independent code in the `workload.rs`. I have created a task for that.

maturar:
**Regarding testing**
We have agreed that we are not going to implement unit tests in this ticket. Current unit tests do not cover how the runtime configuration is mapped into Podman. The mock (Podman test daemon) does not cover entrypoint of the container. Additionally we are going to do massive changes in this area, which will include integration tests. See #1 and #19 . 

I have tested the change manually. I have created own docker image (called "testprintenv") from this ```Dockerfile```:
```
FROM ubuntu:latest
ENTRYPOINT env
```
The point is that this image contains the entrypoint and with the fix from this ticket the Ankaios shall be able to override the entrypoint.
I have created two configurations (startConfig.yaml) with two different runtime configs:

1. 
``` 
    runtimeConfig: |
      image: localhost:5000/testprintenv:latest
```
2. 
```
    runtimeConfig: |
      image: localhost:5000/testprintenv:latest
      command: ["echo"]
      args: ["Hello Ankaios"]
```
[...]

Issue #1, opened 2023-09-05
Support K8S pod spec


windsource:
## Description

For the acceptance of Ankaios some compatability to Kubernetes is important. For that reason the CLI shall be extended to also support (a subset of the) [Kubernetes pod spec](https://kubernetes.io/docs/concepts/workloads/pods/).

## Goals

* The CLI allows to apply a pod spec
* The CLI allows to delete a pod spec which results in stopping the workloads
* The CLI allows to list the pod specs that have been applied
* Podman runtime support is sufficient

## Tasks

- [ ] Define new CLI commands
- [ ] Extend the Ankaios config and PodmanOperator to support K8s pods.
- [ ] CLI understands a pod spec. The CLI shall also pull out the tags of the pod spec and packs them additionally as workload tags.
- [ ] CLI can list pods. 
- [ ] Update SW design

krucod3:
As agreed in the team, we need to check what parts of the pod specification we would like to support (and what is supported by Podman). We also need to check if we would like to also support deployments, configMaps and persistent volume claims.

krucod3:
[Podman supports Kubernetes manifests](https://docs.podman.io/en/latest/markdown/podman-kube-play.1.html) via the `podman play kube` command. Currently the following K8s resources are supported (according to the documentation liked above):
* Pod:
A Podman pod is created which can include multiple containers. Init containers are also supported.
* Deployment:
Unfortunately the replica count does *not* work as expected. It is always overwritten to 1 which kind of degrades the deployment to a pod resources. It is still helpful in case you have a deployment file.
* PersistentVolumeClaim:
Volumes in general works for Pods and Deployments. If you use a PVC, a Podman named volume is created. By default the volumes created with the PVC is not delated when tearing down the Pod. This can still be forced with the --force option.
* Secret:
Creates a Podman named secret which is persisted and can also be used by other Pods. The secret is deleted when tearing down the Pod
* ConfigMap 
The ConfigMap is not persisted in any way by Podman. The data inside is used to fill in variables on the fly while creating the pods in the manifest, but it is gone once the creation process is over. Unlike Secret it does not remain in the system.

Although deployments do not bring a lot of benefit, we should support them to ease up the transition from K8s.
Regarding multiple resources in a manifest, it makes sense to allow ConfigMaps, PVCs and Secrets to be added beside Pod and Deployment. Still there are a couple of questions here:
* What to do if multiple resources creating pods are specified? 
[...]

PR #40, opened 2023-09-11
Allow command entry in Podman RuntimeConfig to overrides image's entrypoint - preparations

maturar:
Issues: #24 

# Definition of Done

- [x] requirements are up-to-date and requirements for new features have been added and mapped to source and test - This will be fixed in the issue #1 
- [x] conform to [unit verification strategy](https://eclipse-ankaios.github.io/ankaios/development/unit-verification/) - This will be fixed in the issue #1

### DIFF
--- a/agent/src/podman/podman_runtime_config.rs
+++ b/agent/src/podman/podman_runtime_config.rs
@@ -1,7 +1,6 @@
 use std::collections::HashMap;
 
 use common::objects::WorkloadSpec;
-use podman_api::models::PortMapping;
 
 use crate::workload_trait::WorkloadError;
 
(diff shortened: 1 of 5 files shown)


agent/src/podman/workload.rs
```diff
@@ -18,94 +18,26 @@ use common::state_change_interface::StateChangeInterface;
 use common::state_change_interface::StateChangeSender;
 
 use hyper::http;
-use podman_api::models::{ContainerMount, PortMapping};
-use podman_api::opts::{ImageListFilter, ImageListOpts, PullOpts};
 use podman_api::Error;
```
krucod3, line 21: We should not have dependencies to podman_api here too.

krucod3 (commented):
Some additional dependencies need to be removed. As I did not know if it would work or not I tried it out and directly committed the changes to the branch.
Please have a look and let me know what you think.

agent/src/podman/podman_runtime_config.rs
```diff
@@ -0,0 +1,68 @@
+use std::collections::HashMap;
+
+use common::objects::WorkloadSpec;
+use podman_api::models::PortMapping;
```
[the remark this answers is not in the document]
maturar: Fixed

agent/src/podman/workload.rs
maturar: Fixed

maturar:
Changes from @krucod3 is okay for me.

krucod3 (approved):
LGTM 👍

Issue #10, opened 2023-09-05
Support volume mounting


windsource:
## Description

For using shared memory communication between the workloads `/dev/shm` must be available. Currently it is not possible to mount this device in the container.

## Goals

- The runtime config has been extended to support mounting of volumes

## Tasks

- [ ] Extend the runtime 
- [ ] Extend the Podman operator

christoph-hamm:
Currently blocked by #24. See [#41](https://github.com/eclipse-ankaios/ankaios/pull/41#issuecomment-1717418128) for more details.

krucod3:
Unblocked, as the preparation parts for #21 - #41 are now on main.

PR #41, opened 2023-09-13
Support volume mounting

christoph-hamm:
Issues: #10

# Definition of Done

- [ ] conform to [unit verification strategy](https://eclipse-ankaios.github.io/ankaios/development/unit-verification/)

### DIFF
--- a/agent/src/podman/mod.rs
+++ b/agent/src/podman/mod.rs
@@ -13,8 +13,9 @@
 // SPDX-License-Identifier: Apache-2.0
 
 mod podman_adapter;
-mod workload;
 mod podman_utils;
+mod runtime_config;
+mod workload;
 pub use podman_adapter::PodmanAdapter;
 
 #[cfg(test)]
--- a/agent/src/podman/runtime_config.rs
+++ b/agent/src/podman/runtime_config.rs
@@ -1,0 +1,123 @@
+use std::collections::HashMap;
+
+use common::objects::WorkloadSpec;
+use podman_api::models::PortMapping;
+
+use crate::workload_trait::WorkloadError;
+
+#[derive(Debug, serde::Deserialize)]
+pub struct PodmanRuntimeConfig {
+    pub image: String,
+    #[serde(default)]
+    pub command: Vec<String>,
+    #[serde(default)]
+    pub args: Vec<String>,
+    #[serde(default)]
+    pub env: HashMap<String, String>,
+    #[serde(default)]
+    pub ports: Vec<Mapping>,
+    #[serde(default)]
+    pub remove: bool,
+    #[serde(default)]
+    pub mounts: Vec<Mount>,
+}
+
+impl PodmanRuntimeConfig {
+    pub fn get_command_with_args(&self) -> Vec<String> {
+        let mut command = vec![];
+        command.extend(self.command.iter().cloned());
+        command.extend(self.args.iter().cloned());
+        command
+    }
+}
+
+impl TryFrom<&WorkloadSpec> for PodmanRuntimeConfig {
+    type Error = TryFromWorkloadSpecError;
+    fn try_from(workload_spec: &WorkloadSpec) -> Result<Self, Self::Error> {
+        match serde_yaml::from_str(workload_spec.workload.runtime_config.as_str()) {
+            Ok(workload_cfg) => Ok(workload_cfg),
+            Err(e) => Err(TryFromWorkloadSpecError(e.to_string())),
+        }
+    }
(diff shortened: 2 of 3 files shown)


christoph-hamm:
Blocked by #24.

Issues #24 move more functionality into  `podman_utils`. This will remove all calls of `podman_api` from `podman/workload.rs`. Hence, it does not make sense to add unit test for this PR until this refactoring is done.

agent/src/podman/podman_runtime_config.rs
```diff
@@ -53,3 +55,56 @@ impl From<TryFromWorkloadSpecError> for WorkloadError {
...
 }
+
+#[derive(Debug, serde::Deserialize)]
+pub struct Mount {
+    #[serde(default)]
+    pub destination: Option<String>,
+    #[serde(default)]
+    pub options: Option<Vec<String>>,
+    #[serde(default)]
+    pub source: Option<String>,
```
krucod3, line 66: We could add some aliases:
```suggestion
    #[serde(default)]
    #[serde(alias = "dst")]
    pub destination: Option<String>,
    #[serde(default)]
    pub options: Option<Vec<String>>,
    #[serde(default)]
    #[serde(alias = "src")]
    pub source: Option<String>,
```

agent/src/podman/podman_runtime_config.rs
```diff
@@ -53,3 +55,56 @@ impl From<TryFromWorkloadSpecError> for WorkloadError {
         WorkloadError::StartError(value.0)
     }
 }
+
+#[derive(Debug, serde::Deserialize)]
+pub struct Mount {
+    #[serde(default)]
+    pub destination: Option<String>,
```
krucod3, line 62: Isn't the destination required in all cases?

krucod3 (changes requested):
I merged with main and tested a bit. Beside the _type, it works, but I have not checked all possible options. 
Some suggestions added.

agent/src/podman/podman_runtime_config.rs
```diff
@@ -53,3 +55,56 @@ impl From<TryFromWorkloadSpecError> for WorkloadError {
...
+    #[serde(default)]
+    pub options: Option<Vec<String>>,
+    #[serde(default)]
+    pub source: Option<String>,
+    #[serde(default)]
+    pub _type: Option<String>,
```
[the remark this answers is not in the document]
christoph-hamm: No, you cannot use the name `type` for an variable, as it is an keyword.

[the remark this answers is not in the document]
krucod3: 🤣 
```suggestion
    #[serde(alias = "type")]
    pub _type: Option<String>,
```

agent/src/podman/podman_runtime_config.rs
christoph-hamm: At the moment, yes destination is always required. Also, I can't image some mount without destination. But, the podman documentation, does not state that destination is always required. It only says that each mount type can have different options and then list these options. It it also an `Option` in podman_api.

agent/src/podman/podman_runtime_config.rs
christoph-hamm: I have added alias for `source` and `destination`. I also have added a rename for `_type`.

agent/src/podman/podman_runtime_config.rs
christoph-hamm: I have now changed `destination` to `String`.

agent/src/podman/podman_runtime_config.rs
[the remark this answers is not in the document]
krucod3: the rename is better 👍

krucod3 (changes requested):
I think we should also mention the new mount option in the reference documentation https://eclipse-ankaios.github.io/ankaios/reference/startup-configuration/#podmanruntimeconfig (`startup-configuration.md`)

maturar:
@christoph-hamm There is a merge conflict here now. It has been caused by the change I made in #24 . Give me a call if you want to know what I have changed in ```podman_utils.rs```. I can help you with resolving the conflict.

christoph-hamm:
> I think we should also mention the new mount option in the reference documentation https://eclipse-ankaios.github.io/ankaios/reference/startup-configuration/#podmanruntimeconfig (`startup-configuration.md`)

I have added the mount option to startup-configuration.md.

krucod3 (approved):
LGTM 👍

PR #50, opened 2023-09-21
Add basic Makefile

isbm:
Add Makefile for:
- Run tests with a faster and better runner
- Check required environment
- Better integration (future) with packaging routines

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,3 +1,10 @@
+[profile.release]
+strip = true
+opt-level = "z"
+lto = true
+codegen-units = 1
+panic = "abort"
+
 [workspace]
 
 members = [
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,20 @@
+.DEFAULT_GOAL := build
+
+.PHONY:build
+release: check-env
+	cargo build -v --release
+
+build: check-env
+	cargo build
+
+test: check-env
+	cargo install cargo-nextest --locked
+	cargo nextest run
+
+check-env:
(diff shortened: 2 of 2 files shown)


lingnoi:
@isbm , firstly, thank you very much for your time and efforts to open a PR to improve Ankaios!

Your suggestion to strip out debug symbols from the binary, 
to use the link time optimization and to define the behavior 
on panicking makes absolute sense to me. 👍 

Also using cargo-nextest makes sense as 
it seems to faster than cargo test (as mentioned here: https://nexte.st/).
Regarding this the follwing questions:
1. Why Makefile is used?
2. Is this also possible to make cargo-nextest available in the devcontainer?
3. Is this also possible to run cargo-nextestin github actions?

lingnoi:
@isbm: Again thank for your suggestions, I discussed your changes with the Ankaios team and came to the following conclusion:
**1. Binary file optimizations:** As Ankaios is still young, don't know yet what kind of optimizations we need (size or speed), so we would leave it unoptimized for now
**2. Debug symbols:** As Ankaios is still young having call stack information on a crash would be nice, so would not strip the binary for now
**3. Using makefile:** for now, we not feel the need for using makefiles and to add an additional tool to our tool pool, but this might change in the future when we feel the need
**4. Using cargo-nextest instead of cargo-test:** This seems to be a great idea, but we would like to have this in our `.devcontainer`. Would you be interested in helping us to bring it in our `.devcontainer`? If yes, this would be awesome. If not, we would create another issue and queue it in our backlog for the next planning.

Issue #55, opened 2023-10-04
Use cargo-nextest instead of cargo-test


lingnoi:
## Description

As suggested in PR #50, use cargo-nextest instead of cargo-test for unit testing because of improved test execution speed.

## Goals

cargo-nextest is used as test runner for unit testing instead cargo-test

## Final result
cargo-test is replaced with cargo-nextest
### Summary

cargo-test shall be replaced with cargo-nextest for unit testing in favor of test execution speed.

### Tasks

- [x] use cargo nextest instead of cargo test #60

PR #60, opened 2023-10-12
replace cargo test with nextest

lingnoi:
Issues: #55 

# Definition of Done

- [x] requirements are up-to-date and requirements for new features have been added and mapped to source and test
- [x] conform to [unit verification strategy](https://eclipse-ankaios.github.io/ankaios/main/development/unit-verification/)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -7,18 +7,21 @@ on:
   workflow_dispatch:
   workflow_call:
 
+env:
+  CARGO_TERM_COLOR: always
+
 jobs:      
   test_and_build_linux_amd64:
     runs-on: ubuntu-latest
     container:
-      image: ghcr.io/eclipse-ankaios/devcontainer-base:0.6.2
+      image: ghcr.io/eclipse-ankaios/devcontainer-base:0.6.3
       options: --privileged
     steps:
       - uses: actions/checkout@v3.5.3
       - uses: Swatinem/rust-cache@v2.6.2
       
       - name: Run tests
-        run: RUST_LOG=debug cargo test -- --show-output --nocapture
+        run: RUST_LOG=debug cargo nextest -- --show-output --nocapture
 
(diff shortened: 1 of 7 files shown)


.github/workflows/build.yml
```diff
@@ -7,18 +7,21 @@ on:
...
-      image: ghcr.io/eclipse-ankaios/devcontainer-base:0.6.2
+      image: ghcr.io/eclipse-ankaios/devcontainer-base:0.6.3
       options: --privileged
     steps:
       - uses: actions/checkout@v3.5.3
       - uses: Swatinem/rust-cache@v2.6.2
       
       - name: Run tests
-        run: RUST_LOG=debug cargo test -- --show-output --nocapture
+        run: RUST_LOG=debug cargo nextest run
```
inf17101, line 24: Does the new test exec lib include the prints/logs by default or why is "nocapture" removed now?

doc/docs/development/run-unit-tests.md
```diff
@@ -14,14 +14,14 @@ This makes difficult to read trace logs, because it is hard to find which trace
 If you want to have traces sorted by unit tests, call this (recommended):
 
 ```shell
-cargo test -- --show-output
+cargo nextest -- --show-output
 ```
 
 Rust also allows to run only a subset of unit tests.
 You have to se the "filter string" in the command:
```
inf17101, line 21:
```suggestion
You have to set the "filter string" in the command:
```

.github/workflows/build.yml
lingnoi: You will get printf/logs outputs when only when a test failed, so you don't need to "pollute" the output with debug logs. Here an example output: 
image

inf17101: Ok I agree.

inf17101 (approved):
LGTM 👍

Issue #75, opened 2023-11-02
Add environment variable for Ankaios address


inf17101:
## Description

Currently if the user changes the Ankaios server socket to another address and/or port the Ankaios user must provide the `--server-address http://<ip>:<port>` all the time when starting an agent and when using the Ankaios CLI. 
This is a bit redundant and can be simplified with adding the environment variable for the Ankaios server socket.

## Goals

Introduce an environment variable for setting the Ankaios Server socket address.

## Final result

When a user uses the Ankaios CLI, the user has now the possibility to set the ANK_SERVER_URL environment variable.

### Summary

Because of the clap issue that the wrong error context is output when an invalid value is set for the environment variable,
we decided to only create an env variable for the server url of the Ankaios CLI. We introduce env variables for the other Ankaios components later when the clap issue is resolved.
I am in discussion with the authors to develop a potential fix in the clap lib itself: https://github.com/clap-rs/clap/issues/5202

### Tasks

- [x] Implement env variable for Ankaios CLI named `ANK_SERVER_URL` that allows to set the server url
- [x] Create requirements

inf17101:
The Ankaios agent and cli use the url (http://127.0.0.1:25551) and the Ankaios server uses the socket address only (127.0.0.1:25551). The question is if we prefer one environment variable and for the agent and the cli we append just the protocol prefix "http://" or we use two different env variables asuming that the three Ankaios components are rarely running on the same system so that it has a benefit to use one env variable.

If we use one env variable then there might be a change necessary if we introduce more protocols, then we can not easily append "http://". Maybe we can use protocol and socket address as two different settings. Then we are flexible if we add more protocols.

What do you think?

windsource:
I think different environment variables for cli, agent and server might be better to have a clear relation between variable and tool but for the way to connect to the server a URL should be used that includes the scheme(protocol like http or https) and the other parts like domain etc.

inf17101:
> I think different environment variables for cli, agent and server might be better to have a clear relation between variable and tool but for the way to connect to the server a URL should be used that includes the scheme (protocol like http or https) and the other parts like domain etc. The URL structure is pretty flexible (see https://en.m.wikipedia.org/wiki/URL) and various schemas have been defined.

I am not sure if three different environment variables have a real benefit. Ankaios CLI and Ankaios agent are using the URL format. So we can have an env variable for the URL format and the socket address for the server.

inf17101:
[...]

PR #79, opened 2023-11-08
Add support for environment variables for server socket and server url

inf17101:
Issues: #75 

# Definition of Done

- [ ] conform to [unit verification strategy](https://eclipse-ankaios.github.io/ankaios/main/development/unit-verification/)

### DIFF
--- a/agent/doc/swdesign/README.md
+++ b/agent/doc/swdesign/README.md
@@ -80,17 +80,48 @@ The following diagram shows the startup sequence of the Ankaios Agent:
 
 Status: approved
 
-The Ankaios Agent shall use the given interfaces and channels to communicate with the Server.
+The Ankaios agent shall use the given interfaces and channels to communicate with the server.
 
 Tags:
 - AgentManager
 
-Rationale: The Server is "only source of true" and ensures that Agents are in the consistent state.
+Rationale: The server is "only source of true" and ensures that Agents are in the consistent state.
 
 Needs:
 - impl
 - itest
 
+#### Agent supports environment variable for the server url
+`swdd~agent-shall-support-environment-variable-for-server-url~1`
+
+Status: approved
+
+When the environment variable with key `ANKAIOS_SERVER_URL` is set and the user has not provided the server url via cli argument,
+the Ankaios agent shall use the server url from the environment variable to connect to the Ankaios server.
+
+Tags:
+- AnkaiosAgent
+
+Needs:
+- impl
+- utest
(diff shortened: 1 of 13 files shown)


inf17101:
PR is not ready yet. I will do a refactoring of the nested conditionals, and add unit tests.

agent/doc/swdesign/README.md
```diff
@@ -80,17 +80,48 @@ The following diagram shows the startup sequence of the Ankaios Agent:
...
 Needs:
 - impl
 - itest
 
+#### Agent supports environment variable for the server url
+`swdd~agent-shall-support-environment-variable-for-server-url~1`
+
+Status: approved
+
+When the environment variable with key `ANKAIOS_SERVER_URL` is set and the user has not provided the server url via cli argument,
```
windsource, line 99: I would prefer, if we cloud derive from the name of an env var the tools it belongs to, e.g.

- `ANK_...` modifies the behavior of the Ankaios cli
- `ANKAGENT_...` modifies the behavior of the Ankaios agent
- `ANKSERVER_...` modifies the behavior of the Ankaios server

So using that notation the variable should be called `ANKAGENT_SERVER_URL`.

windsource: Also I am not sure if we need a separate requirement for every environment variable that we introduce. Just an idea ...

common/src/lib.rs
```diff
@@ -15,7 +15,10 @@
 pub const CHANNEL_CAPACITY: usize = 20;
 pub const DEFAULT_SOCKET_ADDRESS: &str = "127.0.0.1:25551";
 pub const DEFAULT_SERVER_ADDRESS: &str = "http://127.0.0.1:25551";
+pub const SERVER_URL_ENV_KEY: &str = "ANKAIOS_SERVER_URL";
```
windsource, line 18: I would move the keys to the respective files `ank/src/cli.rs`.

windsource:
As discussed the environment variables for the cli are high prio but the ones for agent and server are low prio as the cli is called much more often and server and agent are usually only started at system startup.

agent/doc/swdesign/README.md
windsource: @krucod3 do you agree with having prefixes per component?

inf17101:
I close this PR I had problems with reverting the stuff. I open a new PR. The decisions and discussions are mainly covered in the corresponding issue.

PR #87, opened 2023-11-14
Add cache for podman ps

christoph-hamm:
Issues: #1

Currently for each podman and podman-kube workload the Ankaios agent calls the command `podman ps` once per second. 
This needs a lot CPU resources.
The Ankaios agent shall cache podman ps result, and use the result for all podman workloads.

# Definition of Done


agent/src/runtime_connectors/podman_cli.rs
```diff
@@ -53,6 +55,73 @@ impl From<PodmanContainerInfo> for ExecutionState {
...
+    }
+}
+
+impl Deref for TimedPodmanPsResult {
+    type Target = Mutex<Option<(time::Instant, Arc<PodmanPsResult>)>>;
+
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
```
krucod3, line 91: Note before reading: This is just a question. We don't need to change anything.
We could have also used a type definition, not that it matters. Was there a particular reason to do it via a function?

agent/src/runtime_connectors/podman_cli.rs
```diff
@@ -53,6 +55,73 @@ impl From<PodmanContainerInfo> for ExecutionState {
...
+
+impl From<Result<Vec<PodmanContainerInfo>, String>> for PodmanPsResult {
+    fn from(value: Result<Vec<PodmanContainerInfo>, String>) -> Self {
+        match value {
+            Ok(value) => {
+                let mut container_states = HashMap::new();
+                let mut pod_states: HashMap<String, Vec<ContainerState>> = HashMap::new();
+
+                for i in value {
```
krucod3, line 106: Not the best readability. I've renamed here and there.

agent/src/runtime_connectors/podman_cli.rs
```diff
@@ -212,39 +281,41 @@ impl PodmanCli {
...
+            .as_ref()
+            .pod_states
+            .as_ref()
+            .map_err(|err| err.to_owned())?;
+        Ok(pods
+            .iter()
+            .filter_map(|key| all_pod_states.get(key))
+            .flatten()
+            .map(|x| x.to_owned())
+            .collect())
```
krucod3, line 308: We have a bug here:
If a request is done to ask for the state of 2 pods for a workload and one of these pods is gone, the filter_map function would skip this input entry and return just the container states of the other pod. This would at the end result in a correct running state for the workload.
To reproduce you can create a workload with 2 pods and kill one using podman. The workload is still shown as running although one of its pods is not there anymore.

krucod3 (changes requested):
Finished with the review of the code. There were some small findings that I directly fixed. The fixes are in single commits s.t. they can be reverted if you don't agree with them.

What we definitely have to take care of is the bug with the pod states in case one of the pods in a workloads disappears (see below for details).

I still need to review the tests for the podman_cli.

agent/src/runtime_connectors/podman_cli.rs
```diff
@@ -382,6 +473,8 @@ mod tests {
             state: PodmanContainerState::Created,
             exit_code: 0,
             labels: Default::default(),
+            pod: "".into(),
+            id: "".into(),
```
krucod3, line 477: We could have added a generate_test_podman_container_info function that takes a state and an exit code and returns an object with default values for the rest of the attributes.

agent/src/runtime_connectors/podman_cli.rs
```diff
@@ -616,17 +749,39 @@ mod tests {
...
+                }]
+                .to_json())),
         );
 
         let res = PodmanCli::list_workload_names_by_label("name", "test_agent").await;
         assert_eq!(res, Ok(vec!["workload_name".into()]));
     }
 
+    #[tokio::test]
+    async fn utest_list_workload_names_no_name_label() {
```
krucod3, line 764: Maybe rename the test to:
```suggestion
    async fn utest_list_workload_names_not_found_success() {
```

krucod3 (commented):
Review finished, findings documented.

agent/src/runtime_connectors/podman_cli.rs
christoph-hamm: The case of the whole pod being deleted is easy to fix:

```suggestion
        Ok(pods
            .iter()
            .flat_map(|key| {
                all_pod_states
                    .get(key)
                    .cloned()
                    .unwrap_or_else(|| vec![ContainerState::Unknown])
            })
            .collect())
```

But this will not work if only one container of the pod is removed. In this case the result would still be a "correct running state". I see not how we could fix this, as Podman only returns a list of pods when we create the workload, but not a list of containers.

agent/src/runtime_connectors/podman_cli.rs
christoph-hamm: I have used to Newtype pattern to be able to write an `impl` block for `TimedPodmanPsResult`.

As alternative I could have defined a trait and implement this one for `Mutex<Option<(time::Instant, Arc<PodmanPsResult>)>>`.
I did not want to do this as:

- methods have to be defined twice (once in `trait`, once in `impl`)
- helper methods cannot be defined in trait `impl` blocks

I thinks this Deref implementation does not cause any problems:

- it does not disturb the reading of the implementation of the actual logic.
- if the compiler optimisation is not set to 0, this code should simply disappear.

agent/src/runtime_connectors/podman_cli.rs
christoph-hamm: fixed

server/resources/startConfig.yaml
```diff
@@ -1,60 +1,4 @@
...
-    accessRights:
-      allow: []
-      deny: []
-    tags:
-      - key: owner
-        value: Ankaios team
-    runtimeConfig: |
-      image: alpine:latest
-      commandArgs: [ "echo", "Hello Ankaios"]
   hello-pod:
```
krucod3, line 2: I suppose this was unintentional.

krucod3 (dismissed):
LGTM 👍

krucod3 (approved):
LGTM 👍

agent/src/runtime_connectors/podman_cli.rs
krucod3: Alright

PR #97, opened 2023-11-17
Update documentation on Podman version and quickstart

windsource:
Issues: #1

# Definition of Done

### DIFF
--- a/doc/mkdocs.yml
+++ b/doc/mkdocs.yml
@@ -55,6 +55,7 @@ markdown_extensions:
         - name: mermaid
           class: mermaid
           format: !!python/name:pymdownx.superfences.fence_code_format
+  - pymdownx.details
 
 plugins:
   - social
--- a/doc/mkdocs.yml
+++ b/doc/mkdocs.yml
@@ -55,7 +55,6 @@ markdown_extensions:
         - name: mermaid
           class: mermaid
           format: !!python/name:pymdownx.superfences.fence_code_format
-  - pymdownx.details
 
 plugins:
   - social
--- a/tools/generate_docs.sh
+++ b/tools/generate_docs.sh
@@ -27,9 +27,9 @@ protoc --plugin=protoc-gen-doc="/usr/local/bin/protoc-gen-doc" --doc_out="$targe
 echo "Generate Markdown from ./api/proto/ankaios.proto done."
 
 if [[ "$1" = serve ]]; then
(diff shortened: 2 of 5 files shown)


doc/mkdocs.yml
```diff
@@ -55,6 +55,7 @@ markdown_extensions:
         - name: mermaid
           class: mermaid
           format: !!python/name:pymdownx.superfences.fence_code_format
+  - pymdownx.details
```
krucod3, line 58: Why do we need this extension?

krucod3: Everything works also without it.

tools/generate_docs.sh
```diff
@@ -27,9 +27,9 @@ protoc --plugin=protoc-gen-doc="/usr/local/bin/protoc-gen-doc" --doc_out="$targe
 echo "Generate Markdown from ./api/proto/ankaios.proto done."
 
 if [[ "$1" = serve ]]; then
-    mkdocs serve -f "$target_dir/mkdocs.yml"
+    mkdocs serve --config-file "$target_dir/mkdocs.yml"
```
krucod3, line 30: I tried to serve the documentation locally to check how it works and I'm getting warnings:
```
WARNING -  [10:06:55] "GET /versions.json HTTP/1.1" code 404
```
I searched a bit and found this issue here: https://github.com/squidfunk/mkdocs-material/issues/3164

Should we use `mike` here?

```suggestion
    mike serve --config-file "$target_dir/mkdocs.yml"
```

krucod3 (commented):
Looks good. Only one suggestion.

doc/mkdocs.yml
krucod3:
```suggestion
```

tools/generate_docs.sh
windsource: It's not that easy. `mkdocs serve` and `mike serve` are quite different:

* `mkdocs serve`: Build the documentation and watch for changes and serve the result.
* `mike serve`: Uses an already built documentation which has been committed to the `gh-pages` branch. So it is neither building nor watching.

From my point of view the warning is not that urgent and could be ignored. For preview we do not need the features of mike like switching between different versions.

krucod3: Alright.

krucod3 (approved):
LGMT 👍

PR #102, opened 2023-11-22
Fix new workload always having the state "removed"

christoph-hamm:
If the "podman ps" cache is not reset before starting the state checker, the state checker will get the old "podman ps" result, this result will not contain the new workload, the state checker will assume the workload has be removed and will stop immediately.

# Definition of Done


krucod3 (approved):
LGTM 👍

Issue #84, opened 2023-11-14
Ankaios server should warn if port already used


christoph-hamm:
## Current Behavior

If the Ankaios server port is already in use by another application, the Ankaios server starts up normally and gives not indication that it is not working properly.

## Expected Behavior

The Ankaios server should either crash or at least print a warning, if it is not possible to open the port for the communication with the agents.

## Steps to Reproduce

1. Start some server listening to port 25551, e.g. `python -m http.server 25551`
2. Start the ank-server binary
3. Start the ank-agent binary

## Context (Environment)

Tested on Arch Linux.

### Ankaios server

    [2023-11-14T06:44:06Z INFO  ank_server::ankaios_server] Starting...
    [2023-11-14T06:44:06Z INFO  ank_server::ankaios_server::update_state] The update has 5 new or updated workloads, 0 workloads to delete
    [2023-11-14T06:44:06Z INFO  grpc::execution_command_proxy] Agent agent_A not found, workloads not sent. Waiting for agent to connect.
    [2023-11-14T06:44:06Z INFO  grpc::execution_command_proxy] Agent agent_B not found, workloads not sent. Waiting for agent to connect.

### Ankaios agent

    [2023-11-14T06:44:31Z INFO  ank_agent::agent_manager] Starting ...
    [2023-11-14T06:44:31Z WARN  grpc::client] Connection to server interrupted. Error: Connection interrupted: 'status: Unknown, message: "h2 protocol error: http2 error: connection error detected: frame with invalid size", details: [], metadata: MetadataMap { headers: {} }'
    [2023-11-14T06:44:32Z WARN  grpc::client] Connection to server interrupted. Error: Connection interrupted: 'status: Unknown, message: "h2 protocol error: http2 error: connection error detected: frame with invalid size", details: [], metadata: MetadataMap { headers: {} }'

## Final result

**To be filled by the one closing the issue.**

krucod3:
All done here.

PR #111, opened 2023-11-30
Ankaios server shall exits if the configured gRPC port is already in use

krucod3:
Issues: #84

# Definition of Done


grpc/src/server.rs
```diff
@@ -55,29 +56,31 @@ impl CommunicationsServer for GRPCCommunicationsServer {
...
-                })?;
-            Ok::<(), CommunicationMiddlewareError>(())
-        });
-
-        // TODO these two awaits one after the other do not seem correct ...
-        // [impl->swdd~grpc-server-forwards-commands-to-grpc-client~1]
-        execution_command_proxy::forward_from_ankaios_to_proto(&self.agent_senders, receiver).await;
+                .serve(addr) => {
+                    result.map_err(|err| {
+                        GrpcMiddlewareError::StartError(format!("{err:?}"))
```
inf17101, line 70: Just one hint: the code produces a debug string as error message which contains nested third party error types (tonic, hyper). So the interesting cause is at the very end. I experimented with `err.source()` when bringing `std::error::Error` into scope, but ended up with the same debug string, because tonic TransportError can have many different issues and not all methods are public to do proper checks for interested sub error types.

Shall we keep the debug err message to have a proper fix for the behavior for now and improve the error message when we introduce generic error handling in general? 

image

krucod3: Well, the communication middleware is an exchangeable glue between the server and the agents. In this case the server could not start ("Server startup error") because of the communication middleware ("Communication middleware failure") which in this case is a gRPC middleware ("StartError") which could not start because of a library that is uses.
The current error is maybe long, but it gives some flexibility to exchange parts of the system without too much coupling. The error also is expressive enough as it states where the problem was (although giving too much insides into what libraries we use, which is not a secret).

We could start prescribing how communication middleware error should look like, but we currently support only one and there is a high change that whatever we prescribe will need to be changed later.
As the error message is only logged to the used for reference, I don't see currently the need in any complicated handling here.

inf17101 (approved):
LGTM 👍

Issue #89, opened 2023-11-15
Create smaller devcontainer


inf17101:
## Description

The devcontainer has been grown a lot:
![image](https://github.com/eclipse-ankaios/ankaios/assets/42932060/694402b1-e685-47d6-848b-fc9be01b046b)

This devcontainer is pulled in CI/CD pipeline and also local builds are running very long.

The most dependencies are installed in the Dockerfile.base (the devcontainer base image https://github.com/eclipse-ankaios/ankaios/blob/main/.devcontainer/Dockerfile.base). This is getting to big.

## Goals

Having a slim devcontainer image.

### Summary

**To be filled when the final solution is sketched.**

### Tasks

- [ ] Task 1
- [ ] Task 2
- [ ] ...

inf17101:
I have already started to try to fetch the dependencies to an Alpine image (just to see if it has a big effect when using one of the smallest available images). It does not mean that the final image must built on top of alpine base image.

inf17101:
The debian 11-slim is also a very slim base image (and is not alpine). It has only 80MB.

`docker pull docker.io/debian:11-slim`

![image](https://github.com/eclipse-ankaios/ankaios/assets/42932060/aa2debce-5ab6-4e80-a02b-f3468466500f)

windsource:
I think first we need to analyze the disk usage of the dev container to find out the biggest chunks. `ncdu` is very suited for that.

```shell
ncdu -x /
```

There we can see for `ghcr.io/eclipse-ankaios/devcontainer-base:0.6.4`:

```
--- / -------------------------------------------------------------------------------------------------------------------------------------------------------
    4.0 GiB [######################] /usr
  634.0 MiB [###                   ] /home
   44.6 MiB [                      ] /root
```

and

```
--- /usr ----------------------------------------------------------------------------------------------------------------------------------------------------
                                     /..
    1.7 GiB [######################] /lib
    1.5 GiB [###################   ] /local
  444.3 MiB [#####                 ] /share
  228.2 MiB [##                    ] /bin
```

and

```
--- /usr/lib ------------------------------------------------------------------------------------------------------------------------------------------------
                                     /..
  711.2 MiB [######################] /x86_64-linux-gnu
  360.3 MiB [###########           ] /llvm-14
  183.4 MiB [#####                 ] /jvm
  169.0 MiB [#####                 ] /python3.11
  119.3 MiB [###                   ] /gcc
[...]

PR #116, opened 2023-12-04
Add check for dev containers

koppor:
This PR adds a CI heck for DevContainer builds as a double check if a DevContainer builds.

Maybe, it can even be used as basis to do more build checks - based on the build dev container.

# Definition of Done


.github/workflows/check-dev-containers.yml
```diff
@@ -0,0 +1,49 @@
...
+          - 5000:5000
+    steps:
+      -
+        name: Set up QEMU
+        uses: docker/setup-qemu-action@v2
+      -
+        name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+        with:
+          driver-opts: network=host
```
krucod3, line 30: The qemu and buildx are actually needed to perform a multi-arch build and provide the dev container both for amd64 and arm64. For the verification only I think we can actually spare us these steps in favor of fast builds.

krucod3 (commented):
Reviewed and added some comments.

When we are talking about fast builds, would it be possible to run the workflow only if one of the files changed? Maybe we can take over the checks in the documentation workflow.

inf17101:
> Reviewed and added some comments.
> 
> When we are talking about fast builds, would it be possible to run the workflow only if one of the files changed? Maybe we can take over the checks in the documentation workflow.

@koppor: You could take a look at the `documentation_changes` job inside the documentation workflow and copy and adapt to something similar like the following:
```
...
    - uses: dorny/paths-filter@v2.11.1
      id: filter
      with:
        filters: |
          devconainer:
              - '.devcontainer/**'
...
```

After the change we must also adapt the [self-service](https://eclipse-ankaios.github.io/ankaios/main/development/self-service/) of Eclipse to make the status check required.

.github/workflows/check-dev-containers.yml
```diff
@@ -0,0 +1,49 @@
...
+  containers:
+    runs-on: ubuntu-latest
+    services:
+      registry:
+        image: registry:2
+        ports:
+          - 5000:5000
+    steps:
+      -
+        name: Set up QEMU
```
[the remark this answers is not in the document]
koppor: I learned about that style somewhere and just did copy and paste. 😅

Fixed!

koppor:
> @koppor: You could take a look at the `documentation_changes` job inside the documentation workflow and copy and adapt to something similar like the following:

I used the path filtering directly (as outlined at https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore). Then, this action does not run at all if no changes were made. I think, this is even more efficient.

However, I am not sure how this relates to "required" actions to pass. Needs to be tried out... - In case it works, this would also be helpful for automatic dependency updates including an auto merge (future work).

koppor:
I parallelized the check for base and the "real" image, because the "real" image needs an image from the registry (and does not depend on the build base image).

Refs https://github.com/eclipse-ankaios/ankaios/issues/126 (because an enhancement of this workflow will publish a release of the DevContainer upon successful build and probably also update the `Dockerfile`  automatically).

.github/workflows/check-dev-containers.yml
```diff
@@ -0,0 +1,56 @@
...
+  pull_request:
+  workflow_dispatch:
+  workflow_call:
+
+concurrency:
+  group: "${{ github.workflow }}-${{ github.head_ref || github.ref }}"
+  cancel-in-progress: true
+
+jobs:
+  changes:
```
krucod3, line 16: If we are splitting the build steps for performance, why not splitting the checks too?

.github/workflows/check-dev-containers.yml
```diff
@@ -0,0 +1,56 @@
...
+        filters: |
+          devcontainer:
+            - '.devcontainer/**'
+            - .github/workflows/check-dev-containers.yml
+  container_base:
+    needs: changes
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4.1.1
+        if: ${{ needs.changes.outputs.devcontainer }}
```
krucod3, line 37: I did some tests and it seems like specifying the conditions like this without a comparison lets the job run every time. Why not doing it the same way as in the documentation?
```suggestion
        if: ${{ needs.changes.outputs.devcontainer == 'true' || github.ref_name == 'main' || github.ref_type == 'tag'}}
```

koppor: I thought, GitHub's language would be similar "intelligent" than JavaScript ^^.

krucod3: Me too 🤣

koppor:
I currently don't have the time to continue here. Sorry for the efforts caused!

krucod3:
Sorry to hear that 🙁 
Still thanks a lot for the idea, we will take it over in #89

.github/workflows/check-dev-containers.yml
koppor: I unzip that comment for me:

The time for checking is building the action and the checkout. The check itself is fast.

Thus, when moving the changes-check to the respective job, we gain performance in three ways:

1. avoidance of one checkout call
2. avoidance of starting another job by GitHub workflows
3. better maintenance, because cross-job dependencies are harder to maintain then inside-job dependencies (not scientifically proofed statement ^^)

koppor:
I found some time to continue.

I made following changes:

- Check for changes included in the jobs itself
- Added caching for docker layers (improves build speed by approx one minute)
- Added another job for a chained build: In case the base image changes, the final image is build on the current version of the base image in the repository.

koppor:
@windsource Sure. We can pair up, since I do not have your docker registry credentials. Someone with Ankaios registry push permissions should join the team.

Versioning is hard. We all know. Maybe using `development` or `dev` as "moving" tag? If you intend to release manually, for sure, fixed tags are OK. We could even include [GitVersion](https://github.com/GitTools/GitVersion) to automatically set a meaningful SemVer of the Docker containers.

See line 114 (https://github.com/eclipse-ankaios/ankaios/pull/116/files#diff-d81ffa0ff90993ddb0df7ff9678e69ffa70438cace8f1a34804f3d03de55586eR114) that I even changed the reference of the next Dockerfile to use the previously build one.

Since this is a hobby contribution of me, I would really appreciate it if I could discuss with someone than guess what the architects of Ankaios thought with their containers and especially the container versioning, interpret that, build a separate CI/CD pipeline, push to my own Docker profile, propose something, ask to put credentials somewhere and then incorporate feedback etc. -- It's OK for me if this is the only way, then I'll do as soon I have time.

koppor:
@windsource 

> Probably the build would have to be started manually and then a parameter for the tag needs to be passed. First the check should run and finally the base image needs to be pushed

Currently, the build is done on each change of the image. And @inf17101 even proposed to make the check mandatory: https://github.com/eclipse-ankaios/ankaios/pull/116#issuecomment-1850259501. 

One can also trigger builds if a tag is set. I don't know about your versioning policy in detail. I would version the dev containers along with the releases of Ankaios. Thus, at each "tag" of Ankaois, a bulid and push is triggered.

I am happy to contribute, but I need more guidance about your wishes and also credentials for the Docker registry.

windsource:
Hi @koppor, sure, let's discuss the usage of containers in development and CI for Ankaios. So current we have the base container which serves as base of the dev container during development and also for CI builds. So the base container needs to contain all the tools that are also used for CI builds. Currently we build the base container manually. So before a PR with a change in the base container can be merged, the base container needs to be build manually by a committer and pushed to ghcr. In this step we also increase the version number of the container. Reproducible builds are super important for me. For that reason we do not use latest or dev as label. In order to also support Mac with Apple Silicon we create multi-arch container for amd64 and arm64. As those builds take a long time (one arch always needs to be emulated) we usually use Mac with Apple Silicon for the build as Rosetta does the emulation very efficient.
So although this process is manually we expect to find build problems with the container very fast, as every committer usually uses new versions from main quite often and the CI build with every run. Having that said I am thinking of the benefit of automatic check for building the base container. You did quite some work for the new workflow but that needs to be maintained as well.
As mentioned before the manual build and push step for the base container is not optimal. But on the other hand we also do not change the base container that often. So currently we can live we that. What's you opinion?

koppor:
> So currently we can live we that. What's you opinion?

OK. I understand that as vote for a close.

Issue #117, opened 2023-12-04
Workaround for Podman problems with inconsistent Podman DB


krucod3:
As found by @maturar, Podman 3.4 returns error about an inconsistent state when executing `podman ps` shortly after a workload is scheduled. 
To retain the support for Ubuntu 20.04LTS and 22.04LTS (where we only get the old Podman 3.4 version) we need a workaround for that in order to not show `unknown` state for workloads that are already running and have no problems.

## Current Behavior
System shows unknow state for running workloads if Podman returns an error from the DB

## Expected Behavior
No unknown state is shown for running workloads.

## Steps to Reproduce

1. Start the ank-server with the startup configuration. You can use the configuration from the repository, but without hello-pod workload.
2. Start agents
3. Wait for a while and call ank get workloads
4. Delete all workloads step by step.
5. Stop the agents and the server.
6. Repeat previous steps until you see the error. 

## Logs

Here are traces from Ubuntu 20 (amd64) with the issue:
```
[2023-11-29T10:31:18Z DEBUG ank_agent::runtime_connectors::podman::podman_runtime] The workload 'hello2' has been created with id '32af8436ea999e6d61d95ca7080bc9a9723857f637e689632203509fc9d3fbbc'
[2023-11-29T10:31:18Z DEBUG ank_agent::runtime_connectors::podman::podman_runtime] Starting the checker for the workload 'hello2' with id '32af8436ea999e6d61d95ca7080bc9a9723857f637e689632203509fc9d3fbbc'
[2023-11-29T10:31:18Z TRACE ank_agent::runtime_connectors::podman::podman_runtime] Getting the state for the workload '32af8436ea999e6d61d95ca7080bc9a9723857f637e689632203509fc9d3fbbc'
...
[2023-11-29T10:31:18Z WARN  ank_agent::runtime_connectors::podman::podman_runtime] Could not get state of workload 'PodmanWorkloadId { id: "32af8436ea999e6d61d95ca7080bc9a9723857f637e689632203509fc9d3fbbc" }': 'Execution of command failed: Error: container 83db3d688cf5aa825f75ce0c6c514673af66664fd1b50c9f1d907f84888f8b1a does not exist in database: no such container
    '. Returning unknown.
[2023-11-29T10:31:18Z TRACE ank_agent::runtime_connectors::podman::podman_runtime] Returning the state Unknown
...
[...]

PR #120, opened 2023-12-05
Workaround for Podman problems with inconsistent Podman DB

maturar:
Issues: #117 

# Definition of Done

### DIFF
--- a/agent/src/runtime_connectors/podman_cli.rs
+++ b/agent/src/runtime_connectors/podman_cli.rs
@@ -1082,7 +1082,7 @@ mod tests {
     }
 
     #[tokio::test]
-    async fn utest_list_states_by_id_podman_error() {
+    async fn utest_list_states_by_id_podman_error_retry_failed() {
         let _guard = MOCKALL_CONTEXT_SYNC.get_lock_async().await;
         super::CliCommand::reset();
         *super::LAST_PS_RESULT.lock().await = None;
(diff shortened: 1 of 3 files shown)


agent/src/runtime_connectors/podman_cli.rs
```diff
@@ -91,9 +105,15 @@ impl TimedPodmanPsResult {
     }
 
     async fn new_inner() -> PodmanPsCache {
+        let mut res = PodmanCli::list_states_internal().await;
+        if res.is_err() {
+            // This is a workaround for the known issue in podman (podman ps sometimes fails).
+            log::trace!("'podman ps' has returned error - let's retry it.");
+            res = PodmanCli::list_states_internal().await;
+        }
```
krucod3, line 113: Is this sufficient? Could it be that the second try fails again with the same error?
Maybe we can try to retry a couple (e.g., 3) of times in a loop.

agent/src/runtime_connectors/podman_cli.rs
```diff
@@ -91,9 +105,15 @@ impl TimedPodmanPsResult {
     }
 
     async fn new_inner() -> PodmanPsCache {
+        let mut res = PodmanCli::list_states_internal().await;
+        if res.is_err() {
```
krucod3, line 109: Should we check what kind of Podman error this is and act only on the DB problem?

krucod3 (commented):
I wonder how stable the fix is this way. Are you sure that Podman cannot fail twice in a row with the appropriate timing?

agent/src/runtime_connectors/podman_cli.rs
maturar: That is good question. I was considering that. I was testing this implementation on VM and on the target. My experience says that the error does not happen more than once. Since this is just an workaround, I would suggest to keep the implementation simple (as it is here).

krucod3: Point taken.

agent/src/runtime_connectors/podman_cli.rs
maturar: Most likely we could do that, but I do not know if this issue is caused only by the internal DB problem of the Podman. The second thing is that such thing would not be simple to implement. 
Our helper function does not return the exit code. It would mean to change the helper function. Bigger problem is that the `podman ps` does not have documented "exit codes". For us it would be rather guessing which exit code means a problem with DB. To be on the safe side I would suggest to do the retry after each error.

krucod3: Alright 👍

maturar:
> I wonder how stable the fix is this way. Are you sure that Podman cannot fail twice in a row with the appropriate timing?

Answered in the first finding.

krucod3 (approved):
LGTM 👍

Issue #99, opened 2023-11-21
CLI does not wait for the complete state correctly


krucod3:
## Current Behavior
Currently the CLI send a complete state request and waits for **_A_** response. If the response is the complete state everything works correctly. If the timing is bad, a workload state update could be sent to the CLI in-between and the CLI would output an error and exit.

## Expected Behavior
The CLI waits endlessly for the correct response.

## Steps to Reproduce
Very hard to reproduce. A possible way is to update workloads all the time and request `ank get workloads`.
Still, the issue is clearly visible in code.

## Final result

Added wait for complete state in CLI

PR #124, opened 2023-12-08
Added wait for complete state in cli

lingnoi:
Issues: #99

# Definition of Done


ank/src/cli_commands.rs
```diff
@@ -12,7 +12,7 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 
-use std::fmt;
+use std::{fmt, time::SystemTime};
```
krucod3, line 15:
```suggestion
use std::{fmt, time::{Instant, Duration}};
```

ank/src/cli_commands.rs
```diff
@@ -42,6 +42,7 @@ use url::Url;
 use crate::{cli::OutputFormat, output_and_error, output_debug};
 
 const BUFFER_SIZE: usize = 20;
+const WAIT_TIME_MS: u128 = 3000;
```
krucod3, line 45: We could use Duration here:
```suggestion
const WAIT_TIME_MS: Duration = Duration::from_millis(3000);
```

ank/src/cli_commands.rs
```diff
@@ -223,42 +224,52 @@ impl CliCommands {
...
+        let now = SystemTime::now();
+        loop {
+            if let Some(ExecutionCommand::CompleteState(res)) = self.from_server.recv().await {
+                return Ok(res);
+            }
+            match now.elapsed() {
+                Ok(elapsed) if elapsed.as_millis() == timeout_ms => return Err(CliError::ExecutionError(format!("Failed to get complete state in time (timeout = {timeout_ms}ms)."))),
+                Err(err) => return Err(CliError::ExecutionError(format!("Failed to get elapsed time for timeout calculation.\nErr:{err}"))),
+                _ => continue
+            }
```
krucod3, line 247:
```suggestion
            if request_instant.elapsed() > WAIT_TIME_MS{
                return Err(CliError::ExecutionError(format!("Failed to get complete state in time (timeout={WAIT_TIME_MS:?}).")));
            }
```

ank/src/cli_commands.rs
```diff
@@ -223,42 +224,52 @@ impl CliCommands {
...
+
+        // send complete state request to server
+        self.to_server
+            .request_complete_state(RequestCompleteState {
+                request_id: self.cli_name.to_owned(),
+                field_mask: object_field_mask.clone(),
+            })
+            .await.map_err(|err| CliError::ExecutionError(err.to_string()))?;
+
+        let now = SystemTime::now();
```
krucod3, line 238:
```suggestion
        let request_instant= Instant::now();
```

ank/src/cli_commands.rs
```diff
@@ -223,42 +224,52 @@ impl CliCommands {
         let _ = self.task.await;
     }
 
+    async fn wait_for_complete_state(&mut self, object_field_mask: &Vec<String>, timeout_ms: u128) -> Result<Box<CompleteState>, CliError> {
```
krucod3, line 227: Maybe we can just call it `get_complete_state` and don't say anything about waiting. I also would leave it as you initially wrote it with the constant timeout instead of an input argument.

ank/src/cli_commands.rs
```diff
@@ -223,42 +224,52 @@ impl CliCommands {
...
+        // send complete state request to server
+        self.to_server
+            .request_complete_state(RequestCompleteState {
+                request_id: self.cli_name.to_owned(),
+                field_mask: object_field_mask.clone(),
+            })
+            .await.map_err(|err| CliError::ExecutionError(err.to_string()))?;
+
+        let now = SystemTime::now();
+        loop {
```
krucod3, line 239: The loop will work and I do understand why you build in the timeout too. The thing is, it can still be that you never get a message and the timeout will never be triggered (endless wait).
I initially thought that we can endlessly wait as the response should come, but I do like your idea. The thing is it suggest that we will exit in 3 sec if nothing comes and this is now the case.

We could also wrap the `self.from_server.recv().await` in a `tokio::timeout`, but then the timeout will be kind of started again every time you receive an inappropriate message...

What do you think? Any better suggestion?

krucod3 (commented):
Reviewed and added some suggestions.

ank/src/cli_commands.rs
lingnoi: Another suggestion would be to use `try_recv()` instead of `recv()` so that we can get rid of a possible 'endless wait'.

lingnoi: With `try_recv()` did not work as expected as it returns always `TryRecvError::Empty`. Now I solved the problem with the 'endless wait' of `recv()` by using a `loop` within an `async` block that is wrapped in the `tokio::time::timeout`.

ank/src/cli_commands.rs
```diff
@@ -223,42 +224,53 @@ impl CliCommands {
...
+                request_id: self.cli_name.to_owned(),
+                field_mask: object_field_mask.clone(),
+            })
+            .await.map_err(|err| CliError::ExecutionError(err.to_string()))?;
+
+        let poll_complete_state_response = async {
+            loop {
+                match self.from_server.recv().await {
+                    Some(ExecutionCommand::CompleteState(res)) => return res,
+                    Some(_) | None => (),
```
krucod3, line 242: I the `None` case we can actually directly exit, as the channel was closed from the other side:
```suggestion
                    None => return Err("Channel preliminary closed."),
                    _ => (),
```

But this requires also handling the return values differently (see below) ...

ank/src/cli_commands.rs
```diff
@@ -223,42 +224,53 @@ impl CliCommands {
...
+                }
+            }
+        };
+        match tokio::time::timeout(WAIT_TIME_MS, poll_complete_state_response).await {
+            Ok(res) => Ok(res),
```
krucod3, line 247:
```suggestion
            Ok(Ok(res)) => Ok(res),
            Ok(Err(err)) => Err(CliError::ExecutionError(format!("Failed to get complete state: '{}'", err))),
```

krucod3 (commented):
Great idea with extracting to an async block
👍 

Just one last suggestion, we could also make an early exit if case `None` was retured from the `recv()`.

krucod3 (approved):
LGTM 👍

PR #140, opened 2023-12-20
Fix markdown lint in CI builds

windsource:
Prior to this change, we only triggered the markdown lint check in CI when the documentation had been changed. That could cause a situation when a lint violation exists in SW design which is not recognized immediately in CI but only many commits later when someone changes the documentation.

# Definition of Done

### DIFF
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -66,8 +66,7 @@ jobs:
             tools/generate_docs.sh build
   lint:
     name: Lint documentation
-    needs: documentation_changes
-    if: ${{ needs.documentation_changes.outputs.documentation == 'true' }}
+    # No dependencies as we have markdown files in a lot of places
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/documentation.yml
```diff
@@ -66,8 +66,7 @@ jobs:
             tools/generate_docs.sh build
   lint:
     name: Lint documentation
-    needs: documentation_changes
-    if: ${{ needs.documentation_changes.outputs.documentation == 'true' }}
+    # No dependencies as we have markdown files in a lot of places
```
maturar, line 69: The change is okay for me. Do we still need the job `documentation_changes` with this change?

windsource: Yes, I think we still need that as the `deploy` job depends on it.

maturar: You are right. I have overlooked that dependency.

maturar (approved):

Issue #136, opened 2023-12-18
Ankaios server builds and maintains a dependency graph


krucod3:
## Description

The Ankaios server shall maintain a dependency graph for the workload dependencies. It shall detect cycles in the dependency config and reject an invalid config containing cycles.

## Goals

The Ankaios server shall support workload dependencies and reject invalid dependency configuration.

### Tasks

- [x] Ankaios server can detect cycles in the startup state and rejects states with a cycle by exiting with an error containing the - names of the workloads building the cycle.
- [x] Ankaios server can detect cycles introduced by a set state and rejects the state
- [ ] ~~When Ankaios server rejectes a state because of cycles introduced by a set state the server shall respond with a corresponding message (we need the changes from https://github.com/eclipse-ankaios/ankaios/issues/13 for this)~~ This will be done with #13

inf17101:
For dependency graph representation the already existing data structure `CompleteState` inside the Ankaios server is used. There is no need to introduce a separate data structure currently to represent the dependency graph and to search for cycles. This data structure represents the graph already in a adjacency list style that is suitable for directed graph representation. The adjacency list has the form `HashMap<WorkloadName, HashMap<Dependency Workload Name, AddCondition for Dependency>>` (simplified presentation and part of the CompleteState).

There are several algorithms to detect a cycle in a graph:
- Johnson's algorithm https://www.cs.tufts.edu/comp/150GA/homeworks/hw1/Johnson%2075.PDF
- Floyd's tortoise and hare algorithm
- Brent's algorithm
- Breath first search (BFS)
- Depth first search (DFS)

The decision is to go with classical DFS because we have only mid sized graphs for dependency (we do not expect to have a config with 20 Mio workloads...). It is a common algorithm for detecting cycles and very well engineered. With DFS we can at least find one cycle if we make sure that we visit all the nodes. It finds cycles in linear time complexity (O(|V| + |E|) where V are the amount of vertices and E the amount of Edges, which means we must visit all vertices and edges in the worst case, so going through the whole graph ones). On top of that DFS is easy to understand compared to other algorithms.
[...]

PR #156, opened 2024-01-16
Add cyclic detection for workload dependencies and dependency graph management to Ankaios server

inf17101:
Issues: #136 

# Definition of Done


server/src/ankaios_server/update_state.rs
```diff
@@ -1,438 +0,0 @@
...
-            if !current_state.workloads.contains_key(new_wl_name) {
-                added_workloads.push(new_wls.clone());
-            }
-        });
-
-    if added_workloads.is_empty() && deleted_workloads.is_empty() {
-        return None;
-    }
-
-    log::info!(
```
maturar, line 107: Is this trace available somewhere in the new code? It is rather a matter of taste, but I prefer to have an info trace saying something about content of the update (and therefore about what the server is supposed to do).

inf17101: Yes in general I agree, the reason why I have removed the log was because of effort/value estimation. If I keep this log, then the log is also printed on startup state check as well, which feels strange if it contains a cycle and the cycle error message comes after it. If I want to print this only on subsequent updates, then I need to duplicate code or I must put another "let else" to extract the containers out of the enum again, which I personally did not like (then you need unreachable once more...). I will check what I can do else.

inf17101: Hmm I have kept like before the changes that the "update_state" method returns an execution request. But, I think I can return the containers somehow, but not inside an enum because it feels unhandy here and I think an update_state method should anyway not return a command object, which then another object sends. Let me try something out, I will see how I can change this.

inf17101: I have refactored and checked how it looks like with an tuple as return type and the extra step to extract together with the unreachable statement is even omitted. And we can have back the log only for subsequent update. However, I have now seen that I can directly call the cycle check in the start method, because I know how many workloads were added (all inside the first state). But the downside is that I need something like an init or setter in the server state which I wanted to avoid in general because it potentially bypasses some in-variants of the server state or looks weird. But I will discuss that with @krucod3 tomorrow.

maturar: It is also fine to keep it without a trace. Do not make it too complicated.

inf17101: I have found a solution which has not too much effort. I will commit it tomorrow morning.

inf17101: I have commit the solution of changing the return type. Now the log is output again if the checks on the new state have passed.

maturar: The trace looks good 👍

server/src/ankaios_server/cyclic_check.rs
```diff
@@ -0,0 +1,634 @@
...
+/// detect a cycle in the directed graph for inter workload dependencies within a state.
+///
+/// # Arguments
+///
+/// * `state` - The State with workloads representing the directed graph to check for a cycle
+/// * `start_nodes` - Start visiting the graph only for the passed workloads
+///                   if [`None`] the search is started from all workloads of the state
+///
+pub fn dfs(state: &State, start_nodes: Option<Vec<&String>>) -> Option<String> {
+    // stack is used to terminate the search properly
```
maturar, line 30: What about saying more about the stack and the "termination"? It is not really obvious (to me) how this works.

server/src/ankaios_server/cyclic_check.rs
```diff
@@ -0,0 +1,634 @@
...
+
+    // used to prevent visiting nodes repeatedly
+    let mut visited: HashSet<&String> = HashSet::with_capacity(state.workloads.len());
+
+    /* although the path container is used for lookups,
+    measurements have shown that it is faster than associative data structure within this code path */
+    let mut path: VecDeque<&String> = VecDeque::with_capacity(state.workloads.len());
+
+    // start visiting workloads in the graph only for a subset of workloads (e.g. in case of a an update) or for all
+    let mut data: Vec<&String> = if let Some(workloads_to_visit) = start_nodes {
```
maturar, line 41: Well, I do not like name of the variable `data`, but I do not have better name for it.

server/src/ankaios_server/cyclic_check.rs
inf17101: Thank you, I have tried to make it more clear by putting a comment.

maturar: It looks good now.

server/src/ankaios_server/server_state.rs
```diff
@@ -0,0 +1,843 @@
...
+            .filter(|workload_spec| workload_spec.agent.eq(agent_name))
+            .collect()
+    }
+
+    pub fn update(
+        &mut self,
+        new_state: CompleteState,
+        update_mask: Vec<String>,
+    ) -> Result<AddedDeletedWorkloads, UpdateStateError> {
+        // [impl->swdd~server-state-rejects-state-with-cyclic-dependencies~1]
```
krucod3, line 214: Will be traced below
```suggestion
```

server/src/ankaios_server/server_state.rs
```diff
@@ -0,0 +1,843 @@
...
+                        start_nodes
+                    );
+
+                    if let Some(workload_part_of_cycle) =
+                        cycle_check::dfs(&new_state.current_state, Some(start_nodes))
+                    {
+                        return Err(UpdateStateError::CycleInDependencies(
+                            workload_part_of_cycle,
+                        ));
+                    }
```
krucod3, line 245: We should trace at the place where the check is done.
```suggestion
                    // [impl->swdd~server-state-rejects-state-with-cyclic-dependencies~1]
                    log::debug!(
                        "Execute cyclic dependency check with start_nodes = {:?}",
                        start_nodes
                    );
                    if let Some(workload_part_of_cycle) =
                        cycle_check::dfs(&new_state.current_state, Some(start_nodes))
                    {
                        return Err(UpdateStateError::CycleInDependencies(
                            workload_part_of_cycle,
                        ));
                    }
```

server/doc/swdesign/README.md
```diff
@@ -446,6 +470,84 @@ Needs:
 - utest
 - itest
 
+#### ServerState rejects state with cycle
+`swdd~server-state-rejects-state-with-cyclic-dependencies~1`
+
+Status: approved
+
+When the ServerState is requested to update its State and the new State has a cycle in the workload dependencies, the server shall reject the new State.
```
krucod3, line 478:
```suggestion
When the ServerState is requested to update its State and the new State has a cycle in the workload dependencies, the server shall reject the new State as invalid.
```

krucod3 (commented):
Just some small suggestions.

inf17101:
> Just some small suggestions.

I have commit all suggestions and resolved conversations.

krucod3 (approved):
Reviewed requirements and documentation.
LGTM 👍

PR #160, opened 2024-01-18
Add URL to connection error message

windsource:
This PR adds the server URL to the error message which is printed by e.g. the `ank` command when no connection the server can be established. This helps the user as a wrong URL might have been provided.

# Definition of Done

### DIFF
--- a/grpc/src/client.rs
+++ b/grpc/src/client.rs
@@ -92,7 +92,7 @@ impl CommunicationsClient for GRPCCommunicationsClient {
                         Err(GrpcMiddlewareError::ServerNotAvailable(err)) => {
                             log::debug!("No connection to the server: '{err}'");
                             return Err(CommunicationMiddlewareError(format!(
-                                "Could not connect to server on {}.",
+                                "Could not connect to Ankaios server on {}.",
                                 self.server_address
                             )));
                         }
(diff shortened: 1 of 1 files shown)


grpc/src/client.rs
```diff
@@ -91,7 +91,10 @@ impl CommunicationsClient for GRPCCommunicationsClient {
                         // [impl->swdd~grpc-client-outputs-error-server-unavailability-for-cli-connection~1]
                         Err(GrpcMiddlewareError::ServerNotAvailable(err)) => {
                             log::debug!("No connection to the server: '{err}'");
-                            return Err(CommunicationMiddlewareError("No connection to the server! Make sure that Ankaios Server is running.".to_string()));
+                            return Err(CommunicationMiddlewareError(format!(
+                                "Could not connect to Ankaios server on {}.",
```
krucod3, line 95:
```suggestion
                                "Could not connect to Ankaios server on '{}'.",
```

windsource: Hm, so currently the error message is:

```
error: Communication middleware failure: 'Could not connect to Ankaios server on http://127.0.0.1:25551/.'
```

With your suggested change it would be:

```
error: Communication middleware failure: 'Could not connect to Ankaios server on 'http://127.0.0.1:25551/'.'
```

That looks strange to me.

krucod3 (approved):
LGTM 👍

grpc/src/client.rs
windsource: The detailed message is already in quotes.

Issue #13, opened 2023-09-05
CLI should report state when running or deleting workloads


windsource:
Reported by @krucod3:

## Description

Currently when a workload fails during startup there is only a log message. This is not very informative when using the cli and does not allow proper health management. Additionally deleting workloads should also deliver a result (workload deleted)

## Goals

We should improve the success/error handling of starting workloads s.t.:
* success:
    * cli outputs success or failure
* error:
    * other agents get a workload failed workload state This shall be sent when something during the creation/setup of the Workload fails inside the spawned task.
    * the cli get and outputs the state to the used

We should inform when a workload was deleted.

As mentioned by @windsource, it would be great if we use a create for the CLI output which also supports Windows. The server and the agent are not targeted at Windows, but a user could require to run the CLI on a Windows machine. **We don't need to test the Windows support for now, but we should "keep the door open" for a later support.**

## Tasks

- [ ] Unit tests 
- [ ] System tests
- [ ] User documentation 
- [ ] Design documentation

krucod3:
We had to add some stop handling to the CLI with #100. We should review if we still need the Goodbye messages once this here is implemented.

krucod3:
Maybe we should also add some responses to the `set state`, e.g., accepted or failed.

krucod3:
I'll check the user and design documentation.

maturar:
I am going to review the implementation and I will eventually do the unit tests too (not sure if I can make it before my vacation next week).

lingnoi:
I'll take care for the system tests.

PR #182, opened 2024-02-06
Ank-server returns response message after UpdateStateRequest

christoph-hamm:
Issues: #13

This PR changes `ank-server` to return a response after it handled an UpdateStateRequest. The response can be either anUpdateStateSuccess or an Error. In case of a success, the response contains a list of Workloads which will be removed or added. This information is necessary for the `ank` command to wait for the workloads actually being stopped/started (see issue #13).

# Definition of Done


server/src/ankaios_server.rs
```diff
@@ -435,6 +476,14 @@ mod tests {
         });
         assert_eq!(from_server_command, expected_from_server_command);
 
+        assert!(matches!(
+            comm_middle_ware_receiver.recv().await.unwrap(),
+            FromServer::Response(Response {
+                request_id,
+                response_content: ResponseContent::UpdateStateSuccess(UpdateStateSuccess { added_workloads, deleted_workloads })
+            }) if request_id == REQUEST_ID_A && added_workloads == vec!["workload A"] && deleted_workloads.is_empty()
+        ));
```
maturar, line 485: What about writing this:

```suggestion
        assert_eq!(
            comm_middle_ware_receiver.recv().await.unwrap(),
            FromServer::Response(Response {
                request_id: REQUEST_ID_A.into(),
                response_content: ResponseContent::UpdateStateSuccess(UpdateStateSuccess {
                    added_workloads: vec!["workload B".into()],
                    deleted_workloads: Vec::new(),
                }),
            })
        );
```

This gives better (more readable output) in case of failure. But it is a matter of personal taste of course :-)

common/src/commands.rs
```diff
@@ -277,63 +344,573 @@ pub struct Stop {}
...
+            .as_mut()
+            .unwrap()
+            .startup_state
+            .as_mut()
+            .unwrap()
+            .workloads
+            .insert(
+                WORKLOAD_NAME_1.into(),
+                proto::Workload {
+                    update_strategy: -1,
```
maturar, line 695: Could you use different item than `update_strategy` here? The `update_strategy` has been removed in another pull request #171 

The same problem is in the next test too.

common/src/commands.rs
```diff
@@ -277,63 +344,573 @@ pub struct Stop {}
...
+
+        proto_request_content
+            .startup_state
+            .as_mut()
+            .unwrap()
+            .workloads
+            .insert(
+                WORKLOAD_NAME_1.into(),
+                proto::Workload {
+                    update_strategy: -1,
```
maturar, line 863: Again, try to use different field here, please. It makes merging easier.

server/src/ankaios_server.rs
```diff
@@ -222,13 +229,36 @@ impl AnkaiosServer {
                                     .send(from_server_command)
                                     .await
                                     .unwrap_or_illegal_state();
+                                self.to_agents
```
maturar, line 232: Only an optional improvement. What about writing here a debug trace about sending update_state_success? It would be helpful if you read logs (and you are trying to find out what is going on).

maturar:
Review is done. Findings documented.

common/src/commands.rs
christoph-hamm: I just took a look at #171, and it seems after this PR we can implement `From<proto:Request>` instead of `TryFrom<proto:Requeset>` for `Request`, which make this whole test useless. I think it is better to first merge #171, and then adapt the tests here.

server/src/ankaios_server.rs
christoph-hamm: I have applied the suggestion, but I had change `"workload B"` to `"workload A"` first.

maturar: Oops. This was my mistake. I was trying if the assertion works (and I forgot to revert the change) :-)

server/src/ankaios_server.rs
christoph-hamm: I have added a log message.

common/src/commands.rs
christoph-hamm: Ok, I was wrong. We still need to implement `TryFrom<proto:Requeset>`, as the `dependencies` added to `Workload` use `AddCondition`, which is an enum. I now use `AddCondition` to test a failed conversion.

maturar (approved):
LGTM 👍

PR #243, opened 2024-04-16
Update version to 0.3.1-pre

windsource:
As we are heading towards 0.3.1 I have update the version to 0.3.1-pre.

# Definition of Done

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -28,7 +28,7 @@ dependencies = [
 
 [[package]]
 name = "ank"
-version = "0.3.0"
+version = "0.3.1-pre"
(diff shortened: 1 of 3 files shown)


krucod3 (commented):
Last time we also changed the version of the RUST example. The question is if we want to do this every time. The python and C++ examples also have no version. The Typescript example is at 1.0.0.

If we agree that the versions are independent, I'll create another PR to update the RUST example to a different version and delete the Cargo.lock as it does not get updates since the build is done inside a container.

krucod3 (approved):
LGTM 👍

PR #244, opened 2024-04-16
Use static release of podman for Vagrant Ubuntu 22.04

windsource:
Before this change the podman ubuntu package has been used which is quite old in Ubuntu 22.04 (v3.4.4) and does not support pod specs. This change uses static images of podman and its dependencies and thus installs podman v4.9.4.

# Definition of Done

### DIFF
--- a/tools/vagrant/ubuntu22.04/init.sh
+++ b/tools/vagrant/ubuntu22.04/init.sh
@@ -1,4 +1,15 @@
 #!/bin/bash
 
 apt update
-apt install -y podman tmux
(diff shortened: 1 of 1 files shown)


krucod3 (commented):
Great that this works 😄
Should we also update the ubuntu20.04 machine?

windsource:
> Should we also update the ubuntu20.04 machine?

I think we should keep the Ubuntu 20.04 with the old podman such that we can test that Ankaios really works with that old podman version.

PR #245, opened 2024-04-16
Update versions of control interface examples

krucod3:
The Control Interface examples do not need an explicit version as they are always fitting the release with which they were delivered.

This PR also removes the package lock files since the examples are build in a container and the package lock cannot be easily updated. As these are only examples we don't necessarily need the lock files.

# Definition of Done


maturar (approved):
LGTM :+1:

Issue #236, opened 2024-04-08
CLI could miss UpdateWorkloadState between update and complete state request


krucod3:
If a workload that is in waiting to start state is deleted, it could be that (with the proper timing), the CLI misses the UpdateWorkloadState and declares the workload as removed. This happens because there is a small window between the update and get complete state where the CLI is not listening for the unsolicited messages.
The problem was only reproducible with release builds installed over the install script and started as systemd units.

## Current Behavior

![image](https://github.com/eclipse-ankaios/ankaios/assets/36224699/6d0544d7-0de8-4f86-b80a-44205f33c49b)

## Expected Behavior

The CLI exits correctly when the workload is deleted.

## Steps to Reproduce

1. Install Ankaios via the install script
2. start server and agent via systemd
3. `ank apply` a workload with a dependency which is not available
4. `ank delete workload` the workload
5. CLI hangs and waits for the missed message

## Context (Environment)

Manjaro Linux 
Ankaios v0.3.0-rc2

## Logs
We checked the logs and the message is sent by the agent and received by the server

## Final result

**To be filled by the one closing the issue.**

krucod3:
#246 is now merged. @christoph-hamm, we can now close this here right?

PR #246, opened 2024-04-17
Fix CLI misses messages

christoph-hamm:
Issues: #236

# Definition of Done


ank/src/cli_commands/server_connection.rs
```diff
@@ -0,0 +1,800 @@
+// Copyright (c) 2023 Elektrobit Automotive GmbH
```
krucod3, line 1:
```suggestion
// Copyright (c) 2024 Elektrobit Automotive GmbH
```

ank/doc/swdesign/README.md
```diff
@@ -756,6 +758,26 @@ Needs:
...
+`swdd~cli-stores-unexpected-message~1`
+
+Status: approved
+
+While the Ankaios CLI waits for a response from the Ankaios Server,
+the Ankaios CLI should store all other received messages.
+
+Rationale:
+The Ank CLI needs these message to watch changed workloads,
+as the first `UpdateWorkloadState` message can already be received while waiting for the response the `UpdateStateRequest` and the `CompleteStateRequest`.
```
krucod3, line 775: Weak requirement, please use `shall`. 
Additionally the problem is not completely described in the rationale. Maybe:
```suggestion
While the Ankaios CLI waits for a response from the Ankaios Server,
the Ankaios CLI shall store all unrelated received messages for later processing.

Rationale:
While communicating with the Ankaios server the ank CLI could already receive unsolicited `UpdateWorkloadState` messages that are needed later during command execution. Storing the messages ensure that they are not missed and the CLI does not block endlessly waiting for them, e.g., while waiting for the end state for an apply command.
```

krucod3 (changes requested):
First part of review. Rust code not reviewed completely.

ank/src/cli_commands/server_connection.rs
```diff
@@ -0,0 +1,800 @@
...
+use url::Url;
+
+const BUFFER_SIZE: usize = 20;
+const WAIT_TIME_MS: Duration = Duration::from_millis(3000);
+
+pub struct ServerConnection {
+    to_server: ToServerSender,
+    from_server: FromServerReceiver,
+    task: tokio::task::JoinHandle<()>,
+    missed_from_server_messages: Vec<FromServer>,
```
krucod3, line 40: The CLI is not going to run in production, but I wonder if we should still protect ourselves against excessive memory usage here by using a ring buffer.
In theory this can only happen if the server does not respond in time and we have a lot of workloads in the system that are constantly changing states.
Using a ring buffer would of course break the functionality. Monitoring the size of the vector and exiting above a certain value would also be a possibility.
Admittedly, this is all very theoretical and will probably never occur. Maybe for now we can just keep it in mind.

ank/src/cli_commands/server_connection.rs
```diff
@@ -0,0 +1,800 @@
...
+        object_field_mask: &Vec<String>,
+    ) -> Result<Box<CompleteState>, ServerConnectionError> {
+        output_debug!(
+            "get_complete_state: object_field_mask={:?} ",
+            object_field_mask
+        );
+
+        let request_id = uuid::Uuid::new_v4().to_string();
+
+        // send complete state request to server
```
krucod3, line 88:
```suggestion
```

ank/src/cli_commands/apply_manifests.rs
```diff
@@ -0,0 +1,202 @@
...
+                if parts.len() > 1 {
+                    let _ = &mut workload_paths
+                        .insert(Path::from(format!("{}.{}", parts[0], parts[1])));
+                }
+            }
+
+            Ok((obj, workload_paths.into_iter().collect()))
+        }
+    }
+}
```
krucod3, line 58:
```suggestion
}

```

ank/src/cli_commands/apply_manifests.rs
```diff
@@ -0,0 +1,202 @@
...
+    prefix: &str,
+) -> Vec<String> {
+    let mut filter_masks = paths
+        .iter()
+        .map(|path| format!("{}.{}", prefix, String::from(path)))
+        .collect::<Vec<String>>();
+    filter_masks.sort();
+    filter_masks.dedup();
+    filter_masks
+}
```
krucod3, line 131:
```suggestion
}

```

ank/doc/swdesign/plantuml/seq_set_state.svg
krucod3, line 1: It is kind of long now, but I also don't have a good idea how to make it shorter. If we split it, the readability suffers. Let's leave as is.

krucod3 (changes requested):

ank/src/cli_commands/server_connection.rs
maturar: What about this: we can implement a check somewhere in a code whether the vector has exceeded a threshold or not and print a warning message. If we see this warning message in a future, we will implement clean solution.

christoph-hamm: I think we should keep it as it is. As it is interactive tool, the user will probably stop it if it takes to long. Also it will be run a normal computer and not embedded devices, hence the memory usage is not a huge concern.

If we do something, I would suggest introducing a timeout. This will not solve the problem directly, as many messages in a short time could still lead to the buffer getting big, but the chances are smaller.

ank/src/cli_commands/server_connection.rs
```diff
@@ -0,0 +1,800 @@
...
+                ..Default::default()
+            },
+        )
+    }
+
+    fn instance_name(workload_name: &str) -> WorkloadInstanceName {
+        format!("{workload_name}.{ID}.{AGENT_A}")
+            .try_into()
+            .unwrap()
+    }
```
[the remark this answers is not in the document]
christoph-hamm: I do not understand want you exactly mean. Do you want to replace the functions above with existing `generate_test_...` functions or do you want to use the `generate_test_...` functions in the functions above?

I would not replace the functions above. All existing `generate_test_...` functions take multiple arguments, which would lead to a lot of repeated code in the test.

Using the `generate_test_...` functions in the functions above, would not simplify the code, as the generating of objects using `..Default::default()` is simple and short. I also think instantiating the objects directly is cleaner, as the `generate_test_...` functions take multiple arguments without clue, which argument is for which field. Using object instantiation, everything is named.

ank/src/cli_commands/server_connection.rs
```diff
@@ -0,0 +1,800 @@
...
+
+    #[derive(Clone)]
+    enum CommunicationSimulatorAction {
+        WillSendMessage(FromServer),
+        WillSendResponse(String, ResponseContent),
+        ExpectReceiveRequest(String, RequestContent),
+    }
+
+    impl CommunicationSimulator {
```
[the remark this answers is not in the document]
christoph-hamm: Moved the struct definition and implementation to start of test section

ank/src/cli_commands/server_connection.rs
```diff
@@ -0,0 +1,800 @@
...
+    #[tokio::test]
+    async fn utest_update_state() {
+        let update_state_success = UpdateStateSuccess {
+            added_workloads: vec![WORKLOAD_NAME_1.into()],
+            deleted_workloads: vec![],
+        };
+
+        let mut sim = CommunicationSimulator::default();
+        sim.expect_receive_request(
+            REQUEST,
```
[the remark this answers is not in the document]
christoph-hamm: I have changed the definition above  to `const REQUEST: &str = "request";`.

ank/src/cli_commands/server_connection.rs
```diff
@@ -0,0 +1,800 @@
...
+            .await;
+        assert!(result.is_err());
+        checker.check_communication();
+    }
+
+    // [utest->swdd~cli-stores-unexpected-message~1]
+    #[tokio::test]
+    async fn utest_get_complete_state_other_response_in_between() {
+        let other_response = FromServer::Response(Response {
+            request_id: "other_request".into(),
```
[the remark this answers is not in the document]
christoph-hamm: This is the response to another request. I think "other_request" is correct here.

I have introduced `const OTHER_REQUEST: &str = "other_request";`.

ank/src/cli_commands/server_connection.rs
```diff
@@ -0,0 +1,800 @@
...
+            REQUEST,
+            ResponseContent::CompleteState(Box::new(complete_state_1())),
+        );
+        let (checker, mut server_connection) = sim.create_server_connection();
+
+        let result = server_connection
+            .get_complete_state(&vec![FIELD_MASK.into()])
+            .await;
+        assert!(result.is_ok());
+        assert_eq!(*result.unwrap(), complete_state_1());
```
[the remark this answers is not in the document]
christoph-hamm: It is not possible to compare `Result<Box<CompleteState>, ServerConnectionError>`, as `ServerConnectionError` does not implement `PartialEq`.

ank/src/cli_commands/server_connection.rs
maturar: Maybe you are right, we can keep it as it is. This is also one of the options from Kaloyan. The last suggestion: what about writing here a comment with warning that we have here potential issue? But it is not really necessary.

maturar:
There are some warnings in the server_connection.rs if you run `cargo test`. @christoph-hamm could you fix it, please?

ank/src/cli_commands/server_connection.rs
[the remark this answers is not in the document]
maturar: I have another solution. I have replaced these functions:

```rust
fn complete_state_1() -> CompleteState {
        CompleteState {
            desired_state: State {
                workloads: [workload(WORKLOAD_NAME_1)].into(),
                ..Default::default()
            },
            ..Default::default()
        }
    }

    fn complete_state_2() -> CompleteState {
        CompleteState {
            desired_state: State {
                workloads: [workload(WORKLOAD_NAME_2)].into(),
                ..Default::default()
            },
            ..Default::default()
        }
    }

    fn workload(workload_name: &str) -> (String, StoredWorkloadSpec) {
        (
            workload_name.into(),
            StoredWorkloadSpec {
                agent: AGENT_A.into(),
                runtime: RUNTIME.into(),
                ..Default::default()
            },
        )
    }
```

With one function:

```rust
    fn complete_state(workload_name: &str) -> CompleteState {
        CompleteState {
            desired_state: State {
                workloads: [(
                    workload_name.into(),
                    StoredWorkloadSpec {
                        agent: AGENT_A.into(),
                        runtime: RUNTIME.into(),
                        ..Default::default()
                    },
                )]
                .into(),
                ..Default::default()
            },
            ..Default::default()
        }
    }
```

What do you say to this solution?

maturar:
There is a merge conflict in this PR. @christoph-hamm could you resolve it, please?

ank/src/cli_commands/server_connection.rs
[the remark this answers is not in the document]
christoph-hamm: I have implemented to solution.

christoph-hamm:
> There are some warnings in the `server_connection.rs` if you run `cargo test`. @christoph-hamm could you fix it, please?

The compiler sees some parts of the productive code as dead, as we are not testing `ServerConnection::new`. As this function setting up the connection to the external server, I see no point in adding a unit test for it. This is something which needs to be checked with the integration/system tests.

ank/src/cli_commands/server_connection.rs
krucod3: I agree, this is a decision we took and the correct place to document it is the design decisions section of the swdd.

krucod3:
@christoph-hamm, @maturar, I now disabled the dead code in case the build is done for tests.

ank/doc/swdesign/README.md
```diff
@@ -53,6 +53,37 @@ Considered alternatives:
...
+* This problem is unlikely to occur:
+  (1) the server take a long time to respond or does not respond at all.
+  (2) the server send many messages to the CLI
+* The CLI is intended for interactive usage.
+  If the response takes to long, the user can terminate the CLI.
+* The CLI is intended for use during development and not for use in production.
+  Crashes are more acceptable here.
+* The CLI will be used on a developer PC and not on embedded devices.
+  Memory constraints are not of the biggest concern.
+
```
krucod3, line 86: I rearranged a bit to fit it to the design decision structure normally used in this and the other documents:

```suggestion
#### The CLI does not limit the size of the missed message buffer
`swdd~cli-does-not-limit-missed-message-buffer-size~1`

Status: approved

While the CLI waits for a response from the Ankaios server,
it stores all other messages received from the Ankaios server in a buffer,
as some messages might be needed after the response is received.
If the Ankaios server takes time to response to the request and many other messages are send,
the buffer can grow without limit.

Rationale:

The buffer is left to grow without limit for the following reasons:

* This problem is unlikely to occur as:
  (1) the server must take a long time to respond or does not respond at all.
  (2) the server must send many messages to the CLI
* The CLI is intended for interactive usage.
  If the response takes to long, the user can terminate the CLI.
* The CLI is intended for use during development and not for use in production.
  Crashes are more acceptable here.
* The CLI will be used on a developer PC and not on embedded devices.
  Memory constraints are not of the biggest concern.
* All alternatives need an additional parameter (size of buffer or timeout)

Considered alternatives:

* Introducing a ring buffer.
  If the buffer grows to large, the oldest messages would be dropped.
  This could lead to important messages being missed.
* The size of the buffer could be limited.
  If the buffer would exceed the limit, the CLI would exit with a failure.
* The CLI could timeout, if the Ankaios server takes to long to respond.
  This would not solve the problem directly, but it would reduce the probability of the problem occurring.

```

krucod3 (approved):
LGTM 👍

Issue #222, opened 2024-04-02
Cannot update workloads if runtime config is unchanged


krucod3:
Updating a workload when the runtime config remains the same is broken on main.
The control interface fifo pipes are apparently (see the logs below) deleted after the create for the new workload and cannot be mounted properly. If the workload remains with the same config, the names of the new and old pipes are the same resulting in deleting the pipes for the workload. 

It is questionable if the workloads needs to be updated at all, when we don't have a change in the config, but even so this must be handled and should not result in a broken workload.

## Current Behavior

Updating a workload fails when the runtime config remains the same.

## Expected Behavior
Updating a workload should work. 

Additional considerations: It could be that an update that does not require a restart is prevented from happening, but this would require some extra handling. We need to see what the effort is.

## Steps to Reproduce

1. Start server and agent with one workload running (of succeeded) at the agent
2. Send an update that does not touch the runtime config, e.g., update the restartPolicy
3. Observe the state of the workload via `ank get workloads`

The workload would go into failing state.

## Context (Environment)

Manjaro Linux
Ankaios build from main using dev container

## Logs

``` console
[2024-04-02T10:12:24Z INFO  ank_agent::runtime_manager] Received a new desired state with '1' added and '1' deleted workloads.
[2024-04-02T10:12:24Z DEBUG ank_agent::runtime_manager] Creating control interface pipes for 'WorkloadSpec { instance_name: WorkloadInstanceName { agent_name: "agent_A", workload_name: "simple", id: "2aef335312f4e87e60b2b2f91befbe556aaf5159fe0e6a1b804e80a95acf1b1e" }, tags: [], dependencies: {}, restart_policy: Always, runtime: "podman", runtime_config: "image: alpine:latest\ncommandArgs: [ \"echo\", \"Hello Ankaios\"]\n" }'
[2024-04-02T10:12:24Z TRACE ank_agent::control_interface::directory] Reusing existing directory '"/tmp/ankaios/agent_A_io/simple.2aef335312f4e87e60b2b2f91befbe556aaf5159fe0e6a1b804e80a95acf1b1e"'
[2024-04-02T10:12:24Z TRACE ank_agent::control_interface::fifo] Reusing existing fifo file '"/tmp/ankaios/agent_A_io/simple.2aef335312f4e87e60b2b2f91befbe556aaf5159fe0e6a1b804e80a95acf1b1e/input"'
[...]

PR #251, opened 2024-04-23
Proper reuse of control interface

lingnoi:
Issues: #222 

# Definition of Done


agent/src/workload.rs
```diff
@@ -81,19 +84,62 @@ impl Workload {
...
+
+                        if !current_api_location
+                            .contains(new_control_interface_location.to_str().unwrap())
+                        {
+                            self.set_control_interface(pipes_channel_context_info);
+                        }
+                    }
+                };
+            }
+        }
```
krucod3, line 131: I'm not sure if the `match` here is making things easy to understand. Maybe we can rewrite as follows:
```suggestion
        if self.control_interface.is_none() {
            self.set_control_interface(pipes_channel_context_info);
        } else if pipes_channel_context_info.is_none {
            self.set_control_interface(None);
        } else if is_control_interface_changed(self.control_interface, pipes_channel_context_info) {
            self.set_control_interface(pipes_channel_context_info);
        }
```

We additionally need the `is_control_interface_changed` function.

Do we also need to unit test the two functions? The logic seems relatively complex, maybe we should pinpoint it with some tests to avoid regressions.

agent/src/workload.rs
```diff
@@ -85,49 +85,51 @@ impl Workload {
...
+                    .unwrap()
+                    .contains(new_location.to_str().unwrap());
+            }
+        }
+    }
+
     fn update_control_interface(
         &mut self,
-        spec: &Option<WorkloadSpec>,
+        // spec: &Option<WorkloadSpec>,
```
krucod3, line 128:
```suggestion
```

agent/src/workload.rs
```diff
@@ -85,49 +85,51 @@ impl Workload {
...
+                        .as_ref()
+                        .unwrap()
+                        .get_run_folder(),
+                );
+
+                return !current
+                    .get_api_location()
+                    .to_str()
+                    .unwrap()
+                    .contains(new_location.to_str().unwrap());
```
krucod3, line 121: Can we use an equal here? I somehow don't have a good feeling with the contains. With all the current implementations, it will work, but it could be a pitfall if we change the ids in a distinct way.
```suggestion
                    .eq(new_location.to_str().unwrap());
```

krucod3 (commented):
Looks good now, just a small suggestion about the comparison.

krucod3 (approved):
LGTM 👍

Issue #187, opened 2024-02-09
Documentation: SVGs hard to read in dark mode


windsource:
There are a couple of SVGs in the documentation, like in [here](https://eclipse-ankaios.github.io/ankaios/0.2/architecture/). If you switch the documentation to dark mode, those SVGs are hard to read.

## Current Behavior
SVGs hard to read in dark mode.

## Expected Behavior
SVGs should be easy to read in light and dark mode.

## Steps to Reproduce

1. Goto https://eclipse-ankaios.github.io/ankaios/0.2/architecture/
2. Use the button left of the search field to switch to dark mode.

## Context (Environment)

https://eclipse-ankaios.github.io/ankaios/0.2/architecture/

## Final result

**To be filled by the one closing the issue.**

simone-weiss:
Sth like 
```
![blub](dark.svg#gh-dark-mode-only)
![blub](light.svg#gh-light-mode-only)
```
should be possible

windsource:
https://github.com/squidfunk/mkdocs-material/discussions/2097 might also help.

maturar:
The fix has been implemented.

PR #253, opened 2024-04-24
Documentation: SVGs hard to read in dark mode

maturar:
Issues: #187 

# Definition of Done


windsource:
Using a white background is fine for me. For the image format I prefer png rather than svg. SVGs are not rendered equally on all platforms. In the past we had some problems with missing fonts which were used in SVGs but not available on certain platforms.
In addition to the white background I would also use a border of 10 (in Draw.io: File - Properties - Border Width).

windsource:
There are still some more mermaid titles on the "Control Interface" page that are hard to read in dark mode. Not sure how that can be fixed.

maturar:
I have investigated the topic with titles in diagrams created by mermaid. This seems to be a bug in mermaid. I found out similar bug ticket about it:
https://github.com/laurent22/joplin/issues/4939
The issue had been closed due to inactivity. There are more issues in are "mermaid and colors in the dark mode". For example:
https://forum.obsidian.md/t/mermaid-diagrams-with-unreadable-text-color/11155/4
https://forum.obsidian.md/t/text-colors-for-mermaid-mindmap-graph-are-unreadable-in-dark-mode/68674

Obviously colors and readability is a pain point in mermaid. The best option would probably be to create the title my markdown outside of mermaid. This would be consistent with other non-mermaid diagrams and (honestly speaking) I do not want to invest more hours and days to this topic.

maturar:
All findings have been fixed. The review can continue.

doc/docs/reference/control-interface.md
```diff
@@ -190,6 +177,8 @@ flowchart TD
     further_processing --> fin
 ```
 
+<figcaption>Read response message via control interface</figcaption>
```
windsource, line 180: I think `<figcaption>` is supposed to be used inside a `<figure>`element. As here it used standalone it is not centered. I have tried to add a `<figure>` element around mermaid and the caption but then the image rendered by mermaid gets very small.

So we have the following options:

1. Have a centered caption but very small images - Not good
2. Have no caption at all - Caption is helpful for some images
3. Replace mermaid with images that are rendered offline - Maybe something for the future.
4. Keep this PR as it is

From all options I think the last one is best for the time being

windsource (approved):
Ok with previous comment.

doc/docs/reference/control-interface.md
maturar: I have added comments to the makrdown sources about it. Is it okay for you?

PR #296, opened 2024-06-14
Remove dependencies to URL crate

krucod3:
This PR removes the dependency to the URL crate as the functionality can be easily achieved also without pulling in 28 dependencies into Ankaios.

# Definition of Done

### DIFF
--- a/grpc/src/client.rs
+++ b/grpc/src/client.rs
@@ -14,26 +14,26 @@
 
 use crate::from_server_proxy;
 use crate::from_server_proxy::GRPCFromServerStreaming;
-use crate::grpc_middleware_error::GrpcMiddlewareError;
-use crate::to_server_proxy;
 use crate::grpc_api::{
     self, agent_connection_client::AgentConnectionClient,
     cli_connection_client::CliConnectionClient, to_server::ToServerEnum, AgentHello,
 };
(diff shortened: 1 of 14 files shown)


grpc/src/client.rs
```diff
@@ -44,24 +44,45 @@ enum ConnectionType {
...
     name: String,
-    server_address: Url,
+    server_address: String,
     connection_type: ConnectionType,
 }
 
+fn verify_address_format(server_address: &String) -> Result<(), CommunicationMiddlewareError> {
+    let re = Regex::new(r"^https?:\/\/.+").unwrap_or_illegal_state();
+    if !re.is_match(server_address) {
+        return Err(CommunicationMiddlewareError(format!(
```
maturar, line 54: Could you double-check the function coverage, please? The coverage report says that this return value (error case) is not covered by any test.

krucod3: I don't see the benefit of adding a utest here. We don't have a complicated algorithm and the regex is fairly simple.
The overall coverage of the file is indeed under our target, but this is caused by the requirements of the functionality which demand itests only. 
If you don't mind, I'll propose to be pragmatic here and leave out the utest as it only would bring maintenance effort without any added benefit.

maturar (approved):
LGTM 👍

Issue #229, opened 2024-04-04
Refactor and improve CLI code


krucod3:
## Description

The `ank` CLI code is currently all located in the cli_command.rs module with now more than 1000 lines of code.
The CLI shall be spitted into smaller modules for better maintainability. 
Additionally the design documentation of the CLI must be updated regarding structural view and the tags of the requirements. Currently there are tags for features, e.g., GetWorkload, instead of components.

## Goals

Improve maintainability of the CLI

### Summary

**To be filled when the final solution is sketched.**

### Tasks

- [x] Refactor the code
- [x] Update SwDD

maturar:
The PR with the implementation has been reviewed and merged into `main`.

PR #299, opened 2024-06-20
 Refactor and improve CLI code

maturar:
Issues: #229 

# Definition of Done


ank/src/main.rs
```diff
@@ -151,7 +153,8 @@ async fn main() {
             None => unreachable!("Unreachable code."),
         },
         cli::Commands::Apply(apply_args) => {
-            if let Err(err) = cmd.apply_manifests(apply_args).await {
+            let handle_apply_args: Box<dyn HandleApplyArgs> = Box::new(apply_args.clone());
+            if let Err(err) = cmd.apply_manifests(apply_args, handle_apply_args).await {
```
krucod3: This part is really strange. I understand why it was needed - to enable partial mock of the arguments, s.t. some part are used as values and the functionality is mocked.

Before writing a useless comment asking for a change that could be impossible, I started trying out if it is possible to separate the data from the function in order to enable the mocking. At the end it worked with some mocking of free utility functions: https://github.com/eclipse-ankaios/ankaios/pull/299/commits/2328eb4625359985adb60df722b466c62aeaa6f0
Sorry for directly messing in the branch, but I wanted to be sure it will work 😉

krucod3 (commented):
The changes look good. 
Unfortunately we indeed see now that some parts of the code were not unit tested.
We have a ticket about the set state which is currently in progress and the tests will be written there. what is left are `wait_list_display` and `workload_table_row`. 
I think writing a couple of tests for the two files should be relatively straight forward, so maybe we can do it directly in the scope here.

maturar:
Changes made by @krucod3 are okay.

maturar:
> The changes look good. Unfortunately we indeed see now that some parts of the code were not unit tested. We have a ticket about the set state which is currently in progress and the tests will be written there. what is left are `wait_list_display` and `workload_table_row`. I think writing a couple of tests for the two files should be relatively straight forward, so maybe we can do it directly in the scope here.

I have added unit tests there. @krucod3 could you check them, please?

krucod3 (approved):
LGTM 👍

Issue #337, opened 2024-08-01
Script for updating Ankaios versions


krucod3:
## Description

With #336 we changed to individual versions for all Ankaios packages. To reduce the risk of forgetting to update a version, we should write a script that updates all components. It would be nice if the script also scans for new components are warns or updates them too.

### Summary

**To be filled when the final solution is sketched.**

### Tasks

- [ ] Task 1
- [ ] Task 2
- [ ] ...

PR #339, opened 2024-08-05
Provide a script to update version of Ankaios

windsource:
Issue-Id: #337

# Definition of Done

### DIFF
--- a/tools/update_version.sh
+++ b/tools/update_version.sh
@@ -1,0 +1,60 @@
+#!/bin/bash
+set -e
+
+script_dir=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+base_dir="$script_dir/.."
+workspace_config="$base_dir/Cargo.toml"
+
+usage() {
+    echo "Usage: $0 [--release] VERSION"
+    echo "Update Ankaios files to VERSION."
+    echo "  --release Official release with assets for download."
+    exit 1
+}
+
+# Initialize variables
+release=0
+version=""
+
+# Parse arguments
+while [[ "$#" -gt 0 ]]; do
+    case $1 in
+        --release) release=1; shift ;;
+        -h|--help) usage ;;
+        *)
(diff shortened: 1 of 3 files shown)


tools/update_version.sh
```diff
@@ -0,0 +1,82 @@
...
+done
+
+# Check if VERSION is set
+if [[ -z "$version" ]]; then
+    echo "Error: VERSION is a mandatory argument."
+    usage
+fi
+
+# Extract all packages from the workspace file
+packages=$(awk '/members *= *\[/{flag=1; next} /\]/{flag=0} flag {gsub(/[" ,]/, ""); print}' "$workspace_config")
```
inf17101, line 62: To be more robust against newlines between package names, I suggest the following small change:
```suggestion
packages=$(awk '/members *= *\[/{flag=1; next} /\]/{flag=0} flag && NF {gsub(/[" ,]/, ""); print}' "$workspace_config")
```

inf17101 (commented):
Review finished. Two small findings.

A hint inside the prepare release documentation that this script shall be executed would be useful: [Prepare Release](https://eclipse-ankaios.github.io/ankaios/main/development/ci-cd-release/#preparing-a-release)

tools/update_version.sh
```diff
@@ -0,0 +1,82 @@
+#!/bin/bash
+
+# Copyright (c) 2023 Elektrobit Automotive GmbH
```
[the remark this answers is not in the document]
windsource: OK.

tools/update_version.sh
windsource: I have tested the existing script and it already works with empty lines between the members. The awk statement will then just output an empty item which will be ignored by the line `for pkg in $packages; do`.
For that reason I think we not need to include this change.

inf17101: Ok, I think we can keep it.

inf17101 (approved):
LTGM👍

Issue #173, opened 2024-01-30
Unify Ankaios CLI


krucod3:
## Description

Before the `ank apply` subcommand (#26) was implemented, all commands required a `-f` flag for reading files provided at the command line. The new `apply` subcommand takes a different approach there to improve usability. 

To not make the usage of the ank CLI confusing, we should take care that the CLI behaves the same for all subcommands and change the old ones to use the new approach.

## Goals

Unify the arguments taken by all ank subcommands to make the usage of the CLI easier and clearer.

## Final result
Now, the ank CLI is unified and `ank set state` is being called in the same manner as `ank apply`.

## Tasks
- [x] Refactor the code
- [x] Write unit tests
- [x] Update system tests
- [x] Update and link SwDD

windsource:
I think the only command that is affected by this would be `ank set state`, or?

HorjuRares:
Currently working on this.

krucod3:
Shifted to 0.5 due to lack of time.

inf17101:
@krucod3, @windsource: I have noticed that we are now having two required positional arguments instead of one for set state after reviewing the PR. One or more object field masks and a new state file. 

Previously, the new state file was an optional argument allowing a user to do the following: Executing `ank set state desiredState` without specifying a new state file and it deletes all workloads.

With the current changes in #340, the new state file is a required positional argument and is not optional anymore. Meaning the `Option` as internal type for the cli argument is useless (and the conditional checks on this type as well).

Secondly, with this change the usability has increased in the success case (no -f and similar to other ank commands), but it has decreased when the user does something wrong with the arguments.

This is because, before this change it was allowed to swap the arguments since the state file arg was a named argument.

Now, if a user accidentally swaps the arguments, then it tries to open the file on a field mask because the last argument is always treated as a file since it is now an positional argument:
![image](https://github.com/user-attachments/assets/ea7f6a7e-a76f-4ce0-9f68-045a04b8d389)

When we improve the error messages in the CLI this challenge comes back to us for the set state command.

Like discussed with @windsource, we bring the new changes to main.

inf17101:
I think another previous feature would not work if we make the argument required: To pipe a CompleteState into stdin.

HorjuRares:
@krucod3 
### Final result
Now, the ank CLI is unified and `ank set state` is being called in the same manner as `ank apply`.

### Tasks
- [ ] Refactor the code
- [ ] Write unit tests
[...]

PR #340, opened 2024-08-06
Unify Ankaios cli

HorjuRares:
Issue-Id: #173


inf17101:
@HorjuRares: It seems like 7 system tests are failing. This needs to be fixed.

ank/src/cli.rs
```diff
@@ -137,7 +137,7 @@ pub enum SetCommands {
         #[arg(required = true)]
         object_field_mask: Vec<String>,
         /// A file containing the new State Object Description in yaml format
-        #[arg(short = 'f', long = "file")]
+        #[arg(required = true)]
         state_object_file: Option<String>,
```
inf17101, line 141: Does not need to be an Option anymore, since it is required.

inf17101 (commented):
Review finished.

@HorjuRares: The Ankaios cli is now unifed for the user. But I see internally that the check if we use a stdin reader or reading from file is still duplicated. It is used inside the ank apply and set state functionality. I am thinking about how this can be done and we will open an issue if it is more effort and not doing it here inside this PR.

inf17101 (commented):
Review done. Some small refactoring suggestions also from a method that is not directly touched inside the PR, but it is good because we anyway work on this code path.

ank/src/cli_commands/set_state.rs
```diff
@@ -117,9 +156,230 @@ impl CliCommands {
 //////////////////////////////////////////////////////////////////////////////
 #[cfg(test)]
 mod tests {
-    use std::io;
+    use super::*;
```
inf17101, line 159: We do not use wildard imports normally. Can you import explicitly?

ank/src/cli_commands/set_state.rs
```diff
@@ -117,9 +156,230 @@ impl CliCommands {
...
+        let temp_obj = process_inputs(io::empty(), &state_object_file)
+            .await
+            .unwrap();
+        println!("{:?}", temp_obj);
+
+        let value: Value = serde_yaml::from_str(SAMPLE_CONFIG).unwrap();
+        let expected_obj = Object::try_from(&value).unwrap();
+        println!("{:?}", expected_obj);
+
+        assert_eq!(temp_obj, expected_obj);
```
inf17101, line 296: The `println!`s can be removed.

ank/src/cli_commands/set_state.rs
```diff
@@ -16,94 +16,133 @@ use common::{
...
+            && mask_parts.starts_with(&workload_level_mask_parts)
         {
-            let stored_workload = StoredWorkloadSpec {
-                agent: "".to_string(),
-                runtime: "".to_string(),
-                runtime_config: "".to_string(),
-                ..Default::default()
-            };
+            complete_state.desired_state.workloads.insert(
+                path.parts()[WORKLOAD_NAME_POSITION].to_string(),
```
inf17101, line 46:
```suggestion
                mask_parts[WORKLOAD_NAME_POSITION].to_string(),
```

ank/src/cli_commands/set_state.rs
```diff
@@ -117,9 +156,230 @@ impl CliCommands {
...
+        };
+        let mut mock_server_connection = MockServerConnection::default();
+        mock_server_connection
+            .expect_update_state()
+            .with(eq(updated_state), eq(update_mask.clone()))
+            .return_once(|_, _| {
+                Ok(UpdateStateSuccess {
+                    added_workloads: vec![],
+                    deleted_workloads: vec![],
+                })
```
inf17101, line 336:
```suggestion
                Ok(UpdateStateSuccess::default())
```

ank/src/cli_commands/set_state.rs
```diff
@@ -117,9 +156,230 @@ impl CliCommands {
...
+            no_wait: true,
+            server_connection: mock_server_connection,
+        };
+
+        let set_state_result = cmd.set_state(update_mask, state_object_file).await;
+        assert!(set_state_result.is_ok());
+    }
+
+    #[tokio::test]
+    async fn utest_set_state_failed() {
```
inf17101, line 350: linkage is missing: `// [utest->swdd~cli-provides-set-desired-state~1]`

In addition, the utest setup can be simplified and I suggest to use a more explicit wrong field mask for the readability:

```rust
        let wrong_update_mask = vec!["desiredState.workloads.notExistingWorkload".to_string()];
        let state_object_file = SAMPLE_CONFIG.to_owned();
        let mut mock_server_connection = MockServerConnection::default();
        mock_server_connection
            .expect_update_state()
            .return_once(|_, _| Ok(UpdateStateSuccess::default()));

        let mut cmd = CliCommands {
            _response_timeout_ms: RESPONSE_TIMEOUT_MS,
            no_wait: true,
            server_connection: mock_server_connection,
        };

        let set_state_result = cmd.set_state(wrong_update_mask, state_object_file).await;
        assert!(set_state_result.is_err());
```

ank/src/cli_commands/set_state.rs
```diff
@@ -54,11 +55,82 @@ fn add_default_workload_spec_per_update_mask(
...
+                read_file_to_string(state_object_file.to_string()).map_err(|error| {
+                    CliError::ExecutionError(format!(
+                        "Could not read the state object file.\nError: {}",
+                        error
+                    ))
+                })?;
+            let value: serde_yaml::Value =
+                serde_yaml::from_str(&state_object_data).map_err(|error| {
+                    CliError::YamlSerialization(format!(
+                        "Could not convert to yaml Value.\nError: {}",
```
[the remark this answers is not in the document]
inf17101: I think I can still see `... to yaml Value...` in the code, but I think we can remove the "Value" and make it equal like in the error message for stdin.

ank/src/cli_commands/set_state.rs
```diff
@@ -67,35 +139,15 @@ impl CliCommands {
...
-                            )))?
-                            .clone(),
-                    )
-                    .map_err(|err| CliError::ExecutionError(err.to_string()))?;
-            }
-            complete_state = complete_state_object.try_into()?;
-        }
+
+        let temp_obj = process_inputs(io::stdin(), &state_object_file).await?;
+        add_default_workload_spec_per_update_mask(&object_field_mask, &mut complete_state);
```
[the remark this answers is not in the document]
inf17101: Here point 4 is still open and I need a reply if you have tested that the utests are failing when some values of the constants inside the function are changed to some different values.

ank/src/cli_commands/set_state.rs
```diff
@@ -16,94 +16,133 @@ use common::{
...
+                    .clone(),
+            )
+            .map_err(|err| CliError::ExecutionError(err.to_string()))?;
+    }
+
+    Ok(complete_state_object)
+}
+
 impl CliCommands {
     pub async fn set_state(
```
inf17101, line 119: I think we need a linkage to this function:
```suggestion
    // [impl->swdd~cli-supports-yaml-to-set-desired-state~1]
    pub async fn set_state(
```

inf17101:
I think if the last findings will be resolved, then we can merge the PR.
@HorjuRares: Could you create a ticket about refactoring the two occurrences of the `match` checking for stdin and file name input to a single place/function?

ank/src/cli_commands/set_state.rs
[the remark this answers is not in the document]
inf17101: Ok thank you

ank/src/cli_commands/set_state.rs
```diff
@@ -16,94 +16,134 @@ use common::{
...
 #[cfg(test)]
 use tests::read_to_string_mock as read_file_to_string;
 
 use super::CliCommands;
 
-fn add_default_workload_spec_per_update_mask(
-    update_mask: &Vec<String>,
-    complete_state: &mut CompleteState,
-) {
+fn create_state_with_default_workload_specs(update_mask: &[String]) -> CompleteState {
```
inf17101, line 31:
```suggestion
// [impl->swdd~cli-supports-yaml-to-set-desired-state~1]
fn create_state_with_default_workload_specs(update_mask: &[String]) -> CompleteState {
```

inf17101:
@HorjuRares: The user documentation must be updated. Could you fix that?

inf17101 (approved):
LGTM👍

ank/src/cli_commands/set_state.rs
```diff
@@ -117,9 +156,207 @@ impl CliCommands {
...
+            "desiredState.workloads.nginx3".to_string(),
+        ];
+
+        let complete_state = create_state_with_default_workload_specs(&update_mask);
+
+        assert!(complete_state.desired_state.workloads.contains_key("nginx"));
+        assert!(complete_state
+            .desired_state
+            .workloads
+            .contains_key("nginx2"));
```
inf17101, line 218: We can make this test more strict to also consider the expected values:
```suggestion
        assert_eq!(
            complete_state.desired_state.workloads.get("nginx"),
            Some(&StoredWorkloadSpec::default())
        );

        assert_eq!(
            complete_state.desired_state.workloads.get("nginx2"),
            Some(&StoredWorkloadSpec::default())
        );
```

PR #345, opened 2024-08-07
Refactor control interface

GabyUnalaq:
# Changes
### Renamed some structs from the control_interface:
| Before | After |
|:----------|:-------------|
| PipesChannelContext | ControlInterface |
| PipesChannelContextError | ControlInterfaceError |
| MockPipesChannelContext | MockControlInterface |
| PipesChannelContextInfo | ControlInterfaceInfo |
| MockPipesChannelContextInfo | MockControlInterfaceInfo |
| PipesChannelTask | ControlInterfaceTask |
| MockPipesChannelTask | MockControlInterfaceTask |

### Moved files related to the control_interface:
| Before | After |
|:----------|:-------------|
| src\control_interface\pipes_channel_context.rs | src\control_interface.rs |
| src\control_interface\pipes_channel_context_info.rs | src\control_interface\control_interface_info.rs |
| src\control_interface\pipes_channel_task.rs | src\control_interface\control_interface_task.rs |

### Got rid of wildcard imports.

# Definition of Done


agent/src/runtime_manager.rs
```diff
@@ -1947,11 +1950,11 @@ mod tests {
             .await;
         let _from_authorizer_context = setup_from_authorizer();
 
-        let pipes_channel_mock = MockPipesChannelContextInfo::new_context();
-        pipes_channel_mock
+        let control_interface_mock = MockControlInterfaceInfo::new_context();
+        control_interface_mock
```
inf17101, line 1954:
```suggestion
        let control_interface_info_mock = MockControlInterfaceInfo::new_context();
        control_interface_info_mock
```

inf17101 (commented):
The renamings are fine. Just one small variable name finding. 

In addition, we shall refactor the duplicated methods "create_control_interface(...)" as well. I think it is not good, that the ControlInterfaceInfo struct contains a method "create_control_interface" in general.

agent/src/control_interface.rs
```diff
@@ -111,14 +135,51 @@ impl PipesChannelContext {
...
 
-impl Drop for PipesChannelContext {
+impl Drop for ControlInterface {
     fn drop(&mut self) {
-        self.abort_pipes_channel_task()
+        self.abort_control_interface_task()
+    }
+}
+
+impl TryFrom<control_interface_info::ControlInterfaceInfo> for ControlInterface {
```
inf17101, line 149: This try_from must be tested somehow directly.

inf17101:
Please test the control interface examples one times.
Check the code coverage if it has decreased.

agent/src/control_interface.rs
inf17101: Removed because the TryFrom implementation cannot be tested, because the ControlInterface::new cannot be intercepted with an context (mixture of mock and not mock).

inf17101:
The ControlInterfaceInfo must be mocked because it is needed in the RuntimeManager where it is constructed within methods. This new call must be intercepted then.

With the new control interface module structure together with mockall implementing a `ControlInterface::from_info` method calling the constructor with the parameters from the Info struct is not testable because a mixture of a mock and not mock inside the same module is not testable. The same is valid for a `try_from` implementation, but anyway a try_from is signaling a type conversion, but we are creating fifo files from an info struct (metadata) which is not type conversion and introduces side effects for the reader.

We decided to refactor to a single method to create an ControlInterface: through its constructor `ControlInterface::new(...)` which is already designed for that. The ControlInterfaceInfo is used as a mock inside the RuntimeManager and also needed there as a mock because its constructor is called within bodies of functions. In general that ControlInterfaceInfo is mocked is making the restructuring more compilcated since it enforces to use also the MockAuthorizer which is a field of the Controlinterface.

With the new structure of the control_interface module there where lots of problems with imports of mocks and public API exports in the module files (public api exports and imports are not allowed inside the same file as the compiler treats it as duplicated imports an throws errors).

Achieved is a common and single interface of creating the control interface in different places where also logging can be handled differently in each client and a new module structure with renamed structures that are now more related to the ControlInterface.

inf17101:
> Please test the control interface examples one times. Check the code coverage if it has decreased.

I have tested the examples and their are working.

windsource (approved):
As far as I can see, it looks fine to me.

inf17101:
> As far as I can see, it looks fine to me.

Ok thank you. I will step once through the swdd requirements. So please wait with the merge.

inf17101:
> > As far as I can see, it looks fine to me.
> 
> Ok thank you. I will step once through the swdd requirements. So please wait with the merge.

@windsource: I have adapted the swdds and fixed also some linkage of existing tests.

Issue #358, opened 2024-08-23
Re-connected Ankaios agent not receiving deleted workload when no added workloads are sent


inf17101:
When an Ankaios agent has assigned one workload and the workload will be deleted during the Ankaios agent is disconnected, then the Agent does not delete this workload once it reconnects to the Ankaios server.

## Current Behavior

A reconnected agent with only one workload that will be deleted during the agent is disconnected will lead to a running workload that is not anymore in the desired state and the deleted workload will not be deleted. The agent is blind and does nothing with this workload on startup because the Ankaios server does only sent the added workloads and the workload states when an agent connects.

If the agent has assigned more then one workload and other workloads are not deleted during the agent is disconnected then the Agent receives added workloads from the server and the initial resume/replace logic of the agent runs which correctly deletes the deleted workload. The agent enforces correctly the desired state in this situation.

## Expected Behavior

An Ankaios agent shall always enforce the desired state of the Ankaios server and shall not be out of sync.

## Steps to Reproduce

1. Start an Ankaios server with `server/resources/startConfig.yaml`
2. Start one Ankaios agent with name `agent_A` to make sure that only the nginx workloads runs (the rest of the workloads inside the startConfig can be ignored for this situation)
3. Disconnect the agent after its workload nginx has the running state
4. Execute an `ank -k set state desiredState emptyState.yaml` containing an empty state to delete all workloads or use `ank -k delete workload nginx`
5. Start the `agent_A` again in the other terminal window.
6. Check the logs of server and agent

The agent connects to the server awaiting new commands of it, but since there are no new workloads of that agent inside the desired state, the server sends nothing to this agent.

## Final result

**To be filled by the one closing the issue.**

GabyUnalaq:
Currently working on this.

GabyUnalaq:
After discussions with @krucod3 and @inf17101, the preferred approach will be as follows.

A new message is created, called `ServerHello` (name can be changed) as a response to the `AgentHello`, with information regarding the added workloads of the agent that connected. This way, the functionality remains the same and also this corner-case is covered.

This way, we separate the first response of the server from the regular updates, improving the overall logic of the code and making possible future improvements regarding configuration for the agents.

krucod3:
@GabyUnalaq, PR merged.

PR #367, opened 2024-09-05
Fix reconnected agents not getting deleted workload

GabyUnalaq:
Issue: https://github.com/eclipse-ankaios/ankaios/issues/358

- [x] Create Concept for the fix (detailed in the issue)
- [x] Create ServerHello
- [x] Fix utests and requirements

### DIFF
--- a/agent/src/runtime_manager.rs
+++ b/agent/src/runtime_manager.rs
@@ -106,6 +106,26 @@ impl RuntimeManager {
         }
     }
 
+    pub async fn execute_workloads(
+        &mut self,
+        added_workloads: Vec<WorkloadSpec>,
+        deleted_workloads: Vec<DeletedWorkload>,
+        workload_state_db: &WorkloadStateStore,
+    ) {
+        let workload_operations: Vec<WorkloadOperation> =
+            self.transform_into_workload_operations(added_workloads, deleted_workloads);
+
+        // [impl->swdd~agent-handles-new-workload-operations]
+        // [impl->swdd~agent-handles-workloads-with-fulfilled-dependencies~1]
+        let ready_workload_operations = self
+            .workload_queue
+            .enqueue_filtered_workload_operations(workload_operations, workload_state_db)
+            .await;
+
+        self.execute_workload_operations(ready_workload_operations)
+            .await;
+    }
+
     // [impl->swdd~agent-initial-list-existing-workloads~1]
     pub async fn handle_server_hello(
         &mut self,
@@ -121,17 +141,7 @@ impl RuntimeManager {
             .resume_and_remove_from_added_workloads(added_workloads)
             .await;
 
-        let workload_operations: Vec<WorkloadOperation> =
-            self.transform_into_workload_operations(added_workloads, vec![]);
-
-        // [impl->swdd~agent-handles-new-workload-operations]
-        // [impl->swdd~agent-handles-workloads-with-fulfilled-dependencies~1]
-        let ready_workload_operations = self
-            .workload_queue
-            .enqueue_filtered_workload_operations(workload_operations, workload_state_db)
-            .await;
-
-        self.execute_workload_operations(ready_workload_operations)
+        self.execute_workloads(added_workloads, vec![], workload_state_db)
             .await;
     }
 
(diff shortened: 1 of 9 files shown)


grpc/proto/grpc_api.proto
```diff
@@ -56,9 +56,10 @@ message ToServer {
 */
 message FromServer {
     oneof FromServerEnum {
-        UpdateWorkload updateWorkload = 1; /// A message containing lists of workloads to be added or deleted.
-        UpdateWorkloadState updateWorkloadState = 2; /// A message containing list of workload execution states.
-        ank_base.Response response = 3; /// A message containing a response to a previous request.
+        ServerHello serverHello = 1; /// A message containing information about the workloads to be added after the agent connects.
+        UpdateWorkload updateWorkload = 2; /// A message containing lists of workloads to be added or deleted.
+        UpdateWorkloadState updateWorkloadState = 3; /// A message containing list of workload execution states.
+        ank_base.Response response = 4; /// A message containing a response to a previous request.
```
krucod3, line 62: Let's not reorder the indices of the messages. Please place the server hello at the end with a new index.
We are not yet handling compatibility between versions, but it is not a good practice to change the indexing without a valid reason.

grpc/proto/grpc_api.proto
```diff
@@ -77,6 +78,13 @@ message AgentHello {
 message Goodbye {
 }
 
+/**
+* A message providing information about the workloads to be added after the agent connects.
```
krucod3, line 82: Let's be a bit more precise here and describe that this is a response to the agent hello. Afterwards we can say what it provides.

grpc/src/from_server_proxy.rs
```diff
@@ -109,6 +121,41 @@ pub async fn forward_from_ankaios_to_proto(
...
+                log::trace!("Received ServerHello from server: {:?}.", method_obj);
+
+                let agent_name = method_obj.agent_name.unwrap_or(String::default());
+
+                if !agent_senders.get_all_agent_names().contains(&agent_name) {
+                    log::warn!(
+                        "Could not send added workloads to started agent. Unknown agent: '{}'",
+                        agent_name
+                    );
+                    return;
```
krucod3, line 134: The return is not a good idea here and would break the loop handling messages to agents and clis.
What we want to achieve is to not call the code below. Maybe you can use some `if let Some()` instead

agent/src/runtime_manager.rs
```diff
@@ -108,10 +106,39 @@ impl RuntimeManager {
...
+
+        // [impl->swdd~agent-handles-new-workload-operations]
+        // [impl->swdd~agent-handles-workloads-with-fulfilled-dependencies~1]
+        let ready_workload_operations = self
+            .workload_queue
+            .enqueue_filtered_workload_operations(workload_operations, workload_state_db)
+            .await;
+
+        self.execute_workload_operations(ready_workload_operations)
+            .await;
```
krucod3, line 135: We can maybe extract these lines to new function, as we now have duplicated code.

krucod3 (changes requested):

We should also mention the new ServerHello message in the requirements. Please have a look how AgentHello is described in both agent and server for some inspiration.

grpc/proto/grpc_api.proto
GabyUnalaq: The ServerHello makes sense to be first because of the logic behind it. As for the indexes, I can understand the reason and I will put it at the end.

grpc/proto/grpc_api.proto
GabyUnalaq: Done.

grpc/src/from_server_proxy.rs
GabyUnalaq: Fixed this.

agent/src/runtime_manager.rs
GabyUnalaq: Done.

server/doc/swdesign/README.md
```diff
@@ -309,7 +309,7 @@ Needs:
 
 Status: approved
 
-When an Agent connects to the Server, the Ankaios Server shall send a list of all scheduled Workloads to this Agent as `AddedWorkload` message.
+When an Agent connects to the Server, the Ankaios Server shall send a list of all scheduled Workloads to this Agent as `ServerHello` message.
```
krucod3, line 312: We need to change the version of the requirement too.

krucod3 (commented):
Looks good, just a small finding in the requirements changes.

server/doc/swdesign/README.md
GabyUnalaq: Done.

krucod3 (approved):
LGTM 👍

Issue #341, opened 2024-08-06
workload names containing a dot fail when calling ank apply


sboett-dev:
Workload names containing a dot work in the Ankaios init state file. But when calling `ank apply file.yml`, the ank CLI crashes. 

## Current Behavior
`ank --insecure apply test.yml`
```
thread 'main' panicked at ank/src/cli_commands/apply_manifests.rs:99:60:
called `Option::unwrap()` on a `None` value
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

With backtrace:
`RUST_BACKTRACE=full ank --insecure apply test.yml`
```
thread 'main' panicked at ank/src/cli_commands/apply_manifests.rs:99:60:
called `Option::unwrap()` on a `None` value
stack backtrace:
   0:           0x7d929c - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::hefa070d19712f7b2
   1:           0x818638 - core::fmt::write::hb4781d2f656efdf8
   2:           0x7d60b4 - std::io::Write::write_fmt::ha5c18ee5184597ab
   3:           0x7d90a0 - std::sys_common::backtrace::print::h3107d15dfd66456e
   4:           0x7da40c - std::panicking::default_hook::{{closure}}::h9dbd4e1e1cf8eefd
   5:           0x7da0cc - std::panicking::default_hook::h3fc1e403ea888d0b
   6:           0x7da8ec - std::panicking::rust_panic_with_hook::h04fd07d2f7d1a3b9
   7:           0x7da6c4 - std::panicking::begin_panic_handler::{{closure}}::h6333e84bcadb0df3
   8:           0x7d9780 - std::sys_common::backtrace::__rust_end_short_backtrace::h3f77d7423d385ad2
   9:           0x7da464 - rust_begin_unwind
  10:           0x41e0c8 - core::panicking::panic_fmt::h44130785d025a2df
  11:           0x41e148 - core::panicking::panic::h462d608a37dc0f39
  12:           0x41e080 - core::option::unwrap_failed::hf848e62c66d46e54
  13:           0x49591c - ank::cli_commands::apply_manifests::generate_state_obj_and_filter_masks_from_manifests::h830cca5b098c9392
  14:           0x47c880 - ank::main::{{closure}}::h23d12342ef360f8a
  15:           0x4789e4 - tokio::runtime::park::CachedParkThread::block_on::hfe23bfd57f5a7278
[...]

PR #374, opened 2024-09-09
Fix naming issues

HorjuRares:
Issues: #341 

# Definition of Done

### DIFF
--- a/ank/src/cli_commands/apply_manifests.rs
+++ b/ank/src/cli_commands/apply_manifests.rs
@@ -96,7 +96,16 @@ pub fn update_request_obj(
 ) -> Result<(), String> {
     for workload_path in paths.iter() {
         let workload_name = &workload_path.parts()[1];
-        let cur_workload_spec = cur_obj.get(workload_path).unwrap().clone();
+        cur_obj
+            .clone()
+            .check_if_provided_path_exists(workload_path)
+            .map_err(|err| {
+                format!(
+                    "Got error `{}`. This may be caused by improper naming. Ankaios supports names defined by [a-zA-Z0-9_-].",
+                    err
+                )
+            })?;
+        let cur_workload_spec = cur_obj.get(workload_path).unwrap();
         if req_obj.get(workload_path).is_none() {
             let _ = req_obj.set(workload_path, cur_workload_spec.clone());
         } else {
--- a/ank/src/cli_commands/apply_manifests.rs
+++ b/ank/src/cli_commands/apply_manifests.rs
@@ -16,7 +16,7 @@ use super::{CliCommands, InputSourcePair};
 use crate::cli_commands::State;
 use crate::cli_error::CliError;
 use crate::{cli::ApplyArgs, output_debug};
-use common::objects::CompleteState;
+use common::objects::{CompleteState, STR_RE_WORKLOAD};
 use common::state_manipulation::{Object, Path};
 use std::collections::HashSet;
 
@@ -25,6 +25,8 @@ use self::tests::get_input_sources_mock as get_input_sources;
 
 #[cfg(not(test))]
 use super::get_input_sources;
+
+const WORKLOAD_LEVEL: usize = 1;
 
 // [impl->swdd~cli-apply-supports-ankaios-manifest~1]
 pub fn parse_manifest(manifest: &mut InputSourcePair) -> Result<(Object, Vec<Path>), String> {
@@ -95,16 +97,10 @@ pub fn update_request_obj(
     paths: &[Path],
 ) -> Result<(), String> {
     for workload_path in paths.iter() {
-        let workload_name = &workload_path.parts()[1];
-        cur_obj
-            .clone()
-            .check_if_provided_path_exists(workload_path)
-            .map_err(|err| {
-                format!(
-                    "Got error `{}`. This may be caused by improper naming. Ankaios supports names defined by '^[a-zA-Z0-9_-]+[a-zA-Z0-9_-]*$'.",
-                    err
-                )
-            })?;
+        let workload_name = &workload_path.parts()[WORKLOAD_LEVEL];
+        if !cur_obj.check_if_provided_path_exists(workload_path) {
+            return Err(format!("The provided path does not exist! This may be caused by improper naming. Ankaios supports names defined by '{}'", STR_RE_WORKLOAD));
+        }
         let cur_workload_spec = cur_obj.get(workload_path).unwrap();
         if req_obj.get(workload_path).is_none() {
             let _ = req_obj.set(workload_path, cur_workload_spec.clone());
--- a/ank/src/cli_commands/apply_manifests.rs
+++ b/ank/src/cli_commands/apply_manifests.rs
@@ -788,4 +788,109 @@ mod tests {
             .await;
         assert!(apply_result.is_ok());
     }
+
+    #[tokio::test]
+    async fn utest_apply_manifest_invalid_names() {
+        let _guard = crate::test_helper::MOCKALL_CONTEXT_SYNC
+            .get_lock_async()
+            .await;
+
+        let manifest_content = io::Cursor::new(
+            b"apiVersion: \"v0.1\"\nworkloads:
+            simple.manifest1:
+              runtime: podman
+              agent: agent_A
+              runtimeConfig: \"\"
+                ",
+        );
+
+        let mut manifest_data = String::new();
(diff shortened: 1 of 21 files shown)


common/src/objects/state.rs
```diff
@@ -86,10 +86,11 @@ impl State {
             ));
         }
 
-        let re = Regex::new(r"^[a-zA-Z0-9_-]+[a-zA-Z0-9_-]*$").unwrap();
+        let re_workloads = Regex::new(r"^[a-zA-Z0-9_-]+[a-zA-Z0-9_-]*$").unwrap();
```
krucod3, line 89: Why not just specifying r"^[a-zA-Z0-9_-]+$"?

common/src/objects/state.rs
```diff
@@ -102,9 +103,9 @@ impl State {
                     MAX_CHARACTERS_WORKLOAD_NAME
                 ));
             }
-            if !re.is_match(workload_spec.agent.as_str()) {
+            if !re_agent.is_match(workload_spec.agent.as_str()) {
                 return Err(format!(
-                    "Unsupported agent name. Received '{}', expected to have characters in ^[a-zA-Z0-9_-]+[a-zA-Z0-9_-]*$",
+                    "Unsupported agent name. Received '{}', expected to have characters in ^[a-zA-Z0-9_-]*$",
```
krucod3, line 108: To avoid mistakes we should use the same constant here and at the place of creation of the regex.

common/src/objects/state.rs
HorjuRares: Can the workloads name also be just ""?

common/src/objects/state.rs
HorjuRares: This is the same as the regex for the agent name.

HorjuRares: Done.

common/src/objects/state.rs
krucod3: As agreed, "^[a-zA-Z0-9_-]+$" is the minimal regex to match, `workload-Version_1`, but not `workload.Version.1`.

ank/src/cli_commands/apply_manifests.rs
```diff
@@ -96,7 +96,16 @@ pub fn update_request_obj(
 ) -> Result<(), String> {
     for workload_path in paths.iter() {
         let workload_name = &workload_path.parts()[1];
-        let cur_workload_spec = cur_obj.get(workload_path).unwrap().clone();
+        cur_obj
+            .clone()
+            .check_if_provided_path_exists(workload_path)
```
krucod3, line 101: In general I like how you solved the problem, but I have two points:
* We should try avoiding the clone here if we can. The problem is that `check_if_provided_path_exists` takes a `&mut`. See the comment there for an idea.
* I needed some time to understand what this does, as the error string which gives the information is inside the method. As the `check` function here is fairly simple, we could also return a boolean and spare ourselves the result and make an if out of it.

krucod3 (commented):

ank/src/cli_commands/apply_manifests.rs
```diff
@@ -96,7 +96,16 @@ pub fn update_request_obj(
...
+            .map_err(|err| {
+                format!(
+                    "Got error `{}`. This may be caused by improper naming. Ankaios supports names defined by '^[a-zA-Z0-9_-]+[a-zA-Z0-9_-]*$'.",
```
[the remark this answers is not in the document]
HorjuRares: Done.

common/doc/swdesign/README.md
```diff
@@ -332,6 +332,40 @@ Needs:
...
+- impl
+- utest
+- stest
+
+#### Agent naming convention
+`swdd~common-workload-naming-convention~1`
+
+Status: approved
+
+The Common library shall provide functionality for enforcing an agent name that matches <^[a-zA-Z0-9_-]*$>.
```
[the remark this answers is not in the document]
HorjuRares: Done.

common/src/state_manipulation/object.rs
```diff
@@ -271,6 +271,15 @@ impl Object {
         }
         Some(current_obj)
     }
+
+    pub fn check_if_provided_path_exists(&mut self, path: &Path) -> Result<(), String> {
+        if self.get_as_mapping(path).is_none() {
```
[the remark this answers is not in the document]
HorjuRares: Done.

ank/src/cli_commands/apply_manifests.rs
HorjuRares: Done.

common/doc/swdesign/README.md
```diff
@@ -332,6 +332,45 @@ Needs:
 - impl
 - utest
 
+#### Workload naming convention
+`swdd~common-workload-naming-convention~1`
```
krucod3, line 336: We have not traced this and the next requirement in code and in the tests yet.

krucod3 (commented):
Some additional open points:

We should also enforce the agent name restrictions in the agent startup arguments.

Additionally we need to document the new constraints somewhere in our documentation (in `doc/docs`). The naming constraints can be mentioned as notes in complete-state.md and in the quickstart.md. Additionally they should be present in the documentation of the proto file. The agent constraints must also be added in the documentation of the cli parameters of the `ank-agent`.

It would also be great if we also add some hints about the workload names and how DNS can be enabled for workloads. For this part we would probably need a separate file in the reference folder.

common/doc/swdesign/README.md
HorjuRares: Done

doc/docs/usage/quickstart.md
```diff
@@ -122,6 +122,14 @@ delete it from the state again with:
...
 ```
 
+!!! Note
+
+    In case of workload names, the naming convention states that thier names shall:
+    * contain only regular upper and lowercase characters (a-z and A-Z), numbers and the symbols "-" and "_"
+    * have a minimal length of 1 character
+    * have a maximal length of 63 characters
+    Also, agent name shall contain only regular upper and lowercase characters (a-z and A-Z), numbers and the symbols "-" and "_".
+
```
krucod3, line 132: Maybe we can keep the notes shorter: 
```suggestion
!!! Note

    Workload names shall not be longer then 63 symbols and can contain only regular characters, digits, the "-" and "_" symbols.
```

api/proto/ank_base.proto
```diff
@@ -211,6 +211,13 @@ message WorkloadState {
...
 }
 
+/**
+* In case of workload names, the naming convention states that thier names shall:
+*  - contain only regular upper and lowercase characters (a-z and A-Z), numbers and the symbols "-" and "_"
+*  - have a minimal length of 1 character
+*  - have a maximal length of 63 characters
+* Also, agent name shall contain only regular upper and lowercase characters (a-z and A-Z), numbers and the symbols "-" and "_".
+*/
 message WorkloadInstanceName {
```
krucod3, line 221: This object is only sent by Ankaios and never received from a user of the Control Interface. 

The proper place to describe the limitation is a bit hidden and would be the `WorkloadMap` message.

HorjuRares: Done.

HorjuRares: Done.

server/src/ankaios_server.rs
```diff
@@ -252,21 +245,15 @@ impl AnkaiosServer {
 
                         // [impl->swdd~update-desired-state-with-invalid-version~1]
                         // [impl->swdd~update-desired-state-with-missing-version~1]
-                        if !State::is_compatible_format(
-                            &update_state_request.state.desired_state.api_version,
-                        ) {
-                            log::warn!("The CompleteState in the request has wrong format. Received '{}', expected '{}' -> ignoring the request.",
-                                update_state_request.state.desired_state.api_version, State::default().api_version);
+                        // [impl->swdd~common-workload-naming-convention~1]
+                        // [impl->swdd~common-agent-naming-convention~1]
```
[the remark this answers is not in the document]
HorjuRares: Done.

server/doc/swdesign/README.md
```diff
@@ -869,6 +886,8 @@ Needs:
 - impl
 - utest
 
+####
+
```
krucod3, line 890: Please remove:
```suggestion
```

krucod3 (approved):
LGTM 👍

PR #369, opened 2024-09-05
Add fix for system tests to dev container

windsource:
This PR disables app armor for Podman to prevent the issue, that Podman cannot stop containers. In addition [just](https://just.systems) is added which might replace `make`.

# Definition of Done

The PR shall be merged only if all items mentioned in [CONTRIBUTING.md](https://github.com/eclipse-ankaios/ankaios/blob/main/CONTRIBUTING.md#how-to-contribute) have been followed. In case an item is not applicable as described, please provide a short explanation in the description.

- [ ] All steps in [CONTRIBUTING.md](https://github.com/eclipse-ankaios/ankaios/blob/main/CONTRIBUTING.md#how-to-contribute) have been handled

PR #375, opened 2024-09-09
Fix copyright headers

GabyUnalaq:
# Issue:

Some files did not have the copyright header. This PR aims to fix this and also create a script that checks what files do not have the header present. 

# Definition of Done


windsource:
Maybe we can make the CI build fail when an .rs file is found with no copyright header. In #369 I added [just](https://just.systems) to the dev container. So when #369 got merged, maybe we can replace the Makefile by a Justfile. As the script for checking the copyright is pretty small maybe we can put the content directly in the Justfile then.

krucod3:
When looking into the mass of files we missed, I also welcome the idea with the CI/CD checks.

@GabyUnalaq, writing a general purpose script to add the comment will be harder, but for the CI/CD we just need to check a couple of line at the start of each file.

Currently in the system we have:
* *.rs
* *.robot
* *.resource
* *.py
* *.cpp/*.h/*.hpp/*.c
* *.sh
* *.proto
* *.js
* Dockerfile (starting with (`^`))
* Containerfile (starting with (`^`))
* *.md (We currently don't have copy right for our docs and the design ...)

I suppose checking for "Copyright (c) 202" at the first 2 lines (to account, e.g., for markdown style of comments) would be fine.

Regarding the ignoring of certain paths, we can use the top level `.gitignore`

windsource:
Maybe we should start simple. To just check the Rust files an extension to the `Makefile` like:

```
check-copyright:
	test -z "$(shell find . -name '*.rs' ! -path './target/*' -exec sh -c 'grep -q "// Copyright (c)" {} || echo {}' \;)"
```

should work and then this needs to be included similar to https://github.com/eclipse-ankaios/ankaios/blob/165c293037413b27c830a4fffc283e7f32880468/.github/workflows/build.yml#L62-L63

krucod3:
Maybe we should at least put the copyright in the SwDDs too.

krucod3:
The idea to also search for the "202" was to see if at least some year is mentioned too. Admittedly, we would need changes in 6 years 🤣

windsource:
> Maybe we should at least put the copyright in the SwDDs too.

OK, but since we rarely add files to the SWDD, we may not need a check in CI.

GabyUnalaq:
> The idea to also search for the "202" was to see if at least some year is mentioned too. Admittedly, we would need changes in 6 years 🤣

The search could be done with a regex expression. This way, we cover the year as well.

GabyUnalaq:
This should work now.
I added a regex matching for the Copyright header that we can match with any file and also made the script easy to add extensions and also exclude paths.
Currently, as discussed, the script only looks for `*.rs` files.
Also added it into the CI and Makefile.

windsource:
Well, you created a script with 54 lines which does nearly the same as the single line I have proposed above. Can you elaborate about what's the advantage of the separate script?

tools/check_copyright_headers.sh
```diff
@@ -0,0 +1,54 @@
...
+
+COPYRIGHT_REGEX="Copyright[[:space:]]\\(c\\)[[:space:]]20+[0-9]{2}"
+# List of extensions to search for
+FILE_EXT=(
+    "*.rs"
+)
+# Exclude paths from search
+EXCLUDE_PATHS=(
+    "$SEARCH_DIR/target/*"
+)
```
GabyUnalaq, line 14: It's this part. The regex can fit with any file's copyright header, and in the future if we want more files checked and also we need to add more exclusion paths, we can do it easy.

windsource (approved):
LGTM :+1:

PR #383, opened 2024-09-20
Simplify ank cli set state command

christoph-hamm:
Before the ank cli, when executing a "set state" command, only send the parts of the state which have be changed by the update mask. This was probably done for previous problems with optional fields. This is now been removed and ank cli sends the whole state provided by the users, as:

* the implementation did not allow to delete objects from the desired state
* the ank cli should not have to much knowledge about the implementation. E.g. if we would introduce wildcards in the mask, this would needed to be implemented in the server and the ank cli.

# Definition of Done

### DIFF
--- a/ank/src/cli_commands/set_state.rs
+++ b/ank/src/cli_commands/set_state.rs
@@ -12,10 +12,7 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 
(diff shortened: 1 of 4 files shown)


krucod3 (approved):
LGTM 👍

PR #387, opened 2024-09-30
Fix log message regarding ServerHello

GabyUnalaq:
The log message was wrongly mentioning `UpdateWorkload ` instead of `ServerHello`.


windsource (approved):
LGTM 👍

PR #422, opened 2024-12-04
Fix table update in-place

### LINKED ISSUE #420: Result table is not updated in-place
## Current Behavior

Calling `ank apply` or `ank apply -d`. When the terminal is not wide enough, the result table is not updated in-place as with a wide terminal but repeated below the existing table with every update (e.g. spinner).

image

 

## Expected Behavior

The should be updated in-place whether the terminal is wide enough or not.

## Steps to Reproduce

1. Set terminal to a small width
2. Call `ank apply` or `ank apply -d`
3. Watch output

## Context (Environment)

Ankaios v.0.5.0

## Final result

**To be filled by the one closing the issue.**

windsource:
Before this fix, on narrow terminals the table printed by ank apply has not been updated in-place. Instead it has been repeated below the previous table. The reason for that has been the warning output which disturbed the logic of keeping the old table and replacing that in log.rs:104 fn output_update_fn().

Fixes #420

# Definition of Done

### DIFF
--- a/ank/src/cli_commands/get_workloads.rs
+++ b/ank/src/cli_commands/get_workloads.rs
@@ -11,7 +11,6 @@
 // under the License.
 //
 // SPDX-License-Identifier: Apache-2.0
-use crate::output_warn;
 use crate::{cli_error::CliError, output_debug};
 
 use super::cli_table::CliTable;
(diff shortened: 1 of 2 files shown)


inf17101 (approved):
LGTM 👍

PR #423, opened 2024-12-04
Avoid flickering of table output

windsource:
Before this change, a table output created by ank apply was sometimes flickering. This was caused by printing cleanup strings that had been overwritten by the new message in the same println statement. This change uses a mechanism as described in https://stackoverflow.com/a/71453783.

To fix the stests, this change also detects non-interactive use and does not print control characters in that case.

# Definition of Done


inf17101 (approved):
LGTM 👍

Issue #427, opened 2025-01-07
Control Interface authorization of empty field mask fails with wildcard rule


krucod3:
The ControlInterface authorization of empty field mask ("") fails with a wildcard rule "*"

## Current Behavior
If you use the wildcard filterMask in an authorization rule and make a request with an empty field mask, the request is denied.

## Expected Behavior
Such a request should pass the authorization.

## Steps to Reproduce

1. Create an access rule with a wildcard filter mask, i.e., "*"
2. request the complete state with an empty field mask, i.e., ""
3. Observe that the request is denied

## Context (Environment)

Manjaro with Ankaios 0.5.0

## Final result

**To be filled by the one closing the issue.**

krucod3:
Done.

PR #428, opened 2025-01-08
Fix wildcard auth rule for empty filed masks

krucod3:
Issues: #427 

# Definition of Done

### DIFF
--- a/agent/doc/swdesign/README.md
+++ b/agent/doc/swdesign/README.md
@@ -3160,7 +3160,7 @@ Status: approved
 When the Authorizer checks if an individual entry of the update/field mask of a request matches an individual entry of the filter mask of a deny rule, the Authorizer shall consider them matching if all segments of the allow rule's filter mask match the corresponding segments of the request's update/field mask.
 
 Comment:
-An allow rule matches, if it is the same or a prefix of the request's update/field mask. Consequently, when the allow rule consists only of the wildcards symbol "*", all possible update/field mask, including the empty one, match it.
+A deny rule matches, if the request's update/field mask is the same or a prefix of the rule. Consequently, when the allow rule consists only of the wildcards symbol "*", all possible update/field mask, including the empty one, match it.
 
 Tags:
 - Authorizer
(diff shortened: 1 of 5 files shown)


agent/doc/swdesign/README.md
```diff
@@ -3160,7 +3160,7 @@ Status: approved
 When the Authorizer checks if an individual entry of the update/field mask of a request matches an individual entry of the filter mask of a deny rule, the Authorizer shall consider them matching if all segments of the allow rule's filter mask match the corresponding segments of the request's update/field mask.
 
 Comment:
-A deny rule matches, if the request's update/field mask is the same or a prefix of the rule.
+An allow rule matches, if it is the same or a prefix of the request's update/field mask. Consequently, when the allow rule consists only of the wildcards symbol "*", all possible update/field mask, including the empty one, match it.
```
inf17101, line 3163: Shouldn't this be about `deny rules` in the comment matching the deny rules swdd above?

agent/src/control_interface/authorizer.rs
```diff
@@ -77,7 +77,7 @@ impl Authorizer {
                     } else if let (true, reason) = self.allow_read_write_state_rule.matches(&path) {
                         reason
                     } else {
-                        log::debug!(
+                        log::info!(
```
inf17101, line 80: The global access denied/granted log output in the `control_interface_task.rs` calling the authorizer is also still debug. Shall we change this to info, too?

inf17101 (commented):
Review finished.

agent/doc/swdesign/README.md
krucod3: Copy-paste error 🤣 
```suggestion
A deny rule matches, if the request's update/field mask is the same or a prefix of the rule. Consequently, when the allow rule consists only of the wildcards symbol "*", all possible update/field mask, including the empty one, match it.
```

agent/src/control_interface/authorizer.rs
krucod3: This is a good idea. This way also the request itself would be visible not only the rule that does not match and the Id.

inf17101 (approved):
LGTM 👍

Issue #436, opened 2025-01-27
Performance problems in 0.5.0 (recurrent CPU spikes)


krucod3:
Running a 0.5.0 agent with now workloads while monitoring the performance shows frequent CPU spices of up to 89%. 

## Current Behavior
High recurring CPU spikes

## Expected Behavior
No CPU spikes

## Steps to Reproduce

1. Start an 0.5.0 Ankaios cluster with 1 agent 
2. Observe the CPU usage (e.g., with htop)

## Context (Environment)

Manjaro Linux 
Ankaios 0.5.0 (and current build from main)

## Logs
No logs needed

## Additional Information
| version | htop |
| --- | --- |
| ankaios 0.4.1 | <img width="492" alt="Image" src="https://github.com/user-attachments/assets/2d6408b0-8099-46cc-a5cd-3cac7f06b081" /> |
| ankaios 0.5.0 | <img width="479" alt="Image" src="https://github.com/user-attachments/assets/1030803d-0540-459f-97cd-cc81a78324a1" /> |
| ankaios 0.6.0-pre with resource monitoring compiled out | <img width="532" alt="Image" src="https://github.com/user-attachments/assets/9d614c6a-d10e-4550-ae6c-5ffc22d710e5" />|

## Final result

**To be filled by the one closing the issue.**

krucod3:
It seems like the problem is caused by the resource monitoring introduced with #378 as removing the following code fixes the issue:
``` rust
let mut sys = System::new_with_specifics(
            RefreshKind::new()
                .with_cpu(CpuRefreshKind::everything())
                .with_memory(MemoryRefreshKind::everything()),
        );
```

krucod3:
All done here. We will release this in the upcoming v0.5.1.

PR #438, opened 2025-01-28
Fix high resource usage while sampling resource usage

krucod3:
Issues: #436 

# Definition of Done

### DIFF
--- a/agent/src/agent_manager.rs
+++ b/agent/src/agent_manager.rs
@@ -46,13 +46,13 @@ impl ResourceMonitor {
         }
     }
 
-    fn sample_resource_usage(&mut self) -> (f32, u64) {
+    fn sample_resource_usage(&mut self) -> (CpuUsage, FreeMemory) {
         self.sys.refresh_specifics(self.refresh_kind);
 
         let cpu_usage = self.sys.global_cpu_usage();
         let free_memory = self.sys.free_memory();
 
-        (cpu_usage, free_memory)
+        (CpuUsage::new(cpu_usage), FreeMemory { free_memory })
     }
 }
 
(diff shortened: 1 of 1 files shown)


agent/src/agent_manager.rs
```diff
@@ -30,6 +30,32 @@ use crate::workload_state::WorkloadStateReceiver;
 
 const RESOURCE_MEASUREMENT_INTERVAL_TICK: std::time::Duration = tokio::time::Duration::from_secs(2);
 
+struct ResourceMonitor {
```
HorjuRares, line 33: Great idea which shall later enable addition of new measurements.

HorjuRares: LGTM 👍

agent/src/agent_manager.rs
```diff
@@ -30,6 +30,32 @@ use crate::workload_state::WorkloadStateReceiver;
...
+        let refresh_kind = RefreshKind::new()
+            .with_cpu(CpuRefreshKind::new().with_cpu_usage())
+            .with_memory(MemoryRefreshKind::new().with_ram());
+        ResourceMonitor {
+            refresh_kind,
+            sys: System::new_with_specifics(refresh_kind),
+        }
+    }
+
+    fn sample_resource_usage(&mut self) -> (f32, u64) {
```
christoph-hamm, line 49: This functions name should indicate, that cpu usage and free memory are returned. This gives the reader a change to guess, what the `f32` and `u64` values are.

christoph-hamm (changes requested):

agent/src/agent_manager.rs
krucod3: I did this on the first place, but I want to avoid that somebody writes a new function in case more values are needed. This would require more sampling and more resource usage. I'll update in a different way.

krucod3: done.

christoph-hamm (approved):

Issue #393, opened 2024-10-11
Implement files mounted to a workload


inf17101:
Implement the objects in the interface (external + internal) to allow files mounted to a workload at a certain mount point. The file counted shall be taken from the `desiredState.configs` field were configs for workloads can be defined. Ankaios must render the files content.

This is a sub task of #267, please have a look for the detailed prototype for configs and the current implementation.
To solve this issue the template rendering functionality of #267 must be already ready and working.

Example workload supporting files:

```yaml
...
  dashboard_workload:
    ...
    files:
      - mountPoint: "/some/path/to/file.json"
        data: {{c1.json_data_file_content}}
      - mountPoint: "/some/path/to/bin_data"
        binaryData: {{c1.some_data}}
   ...
```

I have already introduced and experimented with my current configs implementation branch how to bring in this new layout of the `files` field in proto format. I achieved that CLI + Server accepts the new format. It's still in draft and contains only the necessary new objects and conversion functions, the rest is draft code. It is commit to separate branch: https://github.com/eclipse-ankaios/ankaios/tree/267_config_files_impl

## Final result

The new feature is named "workload files" finally. A workload configuration with runtime podman supports the new `files` field allowing to define a mount point for the file inside workload and either text-based data or base64 encoded data (e.g. for binaries).

### Summary

For workloads using `podman-kube` the files feature is not supported  and a workload with files configured gets the ExecutionState `Pending(StartingFailed)` with additional information about unsupported feature and rather using the built-in kube feature `ConfigMap` feature instead.

For workloads with runtime `podman` an Ankaios agent creates the workload files with their corresponding content inside the run folder in the specific workload folder (subfolder `files/`) on the host file system and mounts each file in readonly mode to the podman container at the specified mount path. The content of the `binaryData` field assumes base64 encoded data using the base64 standard encoding including padding according to the base64 RFC and the Ankaios agent decodes the base64 to let the container access the decoded content.

The `data` and `binaryData` fields are rendered with the handlebars template engine allowing to separate configuration from the workload itself and updating the files content separately from the workloads.

Tasks:
- [x] Takeover the branch with already brought in new objects https://github.com/eclipse-ankaios/ankaios/tree/267_config_files_impl , merge the current main branch or at least the newest configuration implementation in the configs branches of other PRs
- [x] Call the existing ConfigRenderer to render also the files fields
- [x] Implement the mount of the files in the supported runtimes (podman)
[...]

PR #445, opened 2025-01-31
Add support for auto line indentation of multi line config to ConfigRenderer

inf17101:
The PR fixes expansion of multi-line configuration items with no indentation level consideration when using the default handlebars template syntax.

Scenario:
A user defines a multi-line sharable configuration item in the state and uses this configuration item in a workload field where line indentation must be considered in the current context during expansion of templates. The indentation level shall be considered to ensure validity of indentation-sensitive formats like YAML.

**Changes:**
- register a custom partial named `indent` in the handlebars template engine object (when rendering partials the line indentation will be automatically considered according to the current render context)
- prevent escaping special character sequences like quotes `\"` to `&quot`, ...
- user documentation about sharable configuration with multi-line content

The added system test shows an example use case with podman kube's ConfigMaps where it is important to keep line indentation while expanding templates in workload configuration.

# Definition of Done

### DIFF
--- a/doc/docs/usage/manifest/config-rendering.md
+++ b/doc/docs/usage/manifest/config-rendering.md
@@ -93,7 +93,7 @@ http {
 ...
 ```
 
-By using the `indent` control structure, incorrect indentation will be prevented and the line indentation of the current context will be considered. This results in an error-free YAML file.
+By using the `indent` control structure, the line indentation of the current context will be considered which results in an error-free YAML file.
 
 ```yaml
 ...
--- a/doc/docs/usage/manifest/config-rendering.md
+++ b/doc/docs/usage/manifest/config-rendering.md
@@ -1,0 +1,112 @@
+# Sharable configuration
+
+Ankaios supports the sharable configuration approach, which allows configurations to be defined once and assigned to any number of workloads. It uses the [handlebars](https://github.com/sunng87/handlebars-rust) templating syntax to extend certain fields of workloads. For a detailed look at supported workload fields and a basic example of sharing configuration with workloads, see [here](../../reference/startup-configuration.md).
+
+## Indentation for multi-line configuration
+
+When using the default handlebars template syntax (`{{config_variable}}`), the line indentation of the current context is not considered. To ensure the validity of certain layouts that rely on the indentation level of multi-line configuration, utilize the following custom `indent` control structure, highlighted below:
+
+```yaml linenums="1" hl_lines="16"
+apiVersion: v0.1
+workloads:
+  frontend:
+    agent: agent_A
+    runtime: podman-kube
+    configs:
+      nginx_conf: nginx_config
+    runtimeConfig: |
+      manifest: |
+        apiVersion: v1
+        kind: ConfigMap
+        metadata:
+          name: nginx-config
+        data:
+          nginx.conf: |
+            {{> indent content=nginx_conf}}
+        ---
+        apiVersion: v1
+        kind: Pod
+        metadata:
+          name: nginx-pod
+        spec:
+          restartPolicy: Never
+          containers:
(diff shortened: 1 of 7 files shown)


server/src/ankaios_server/config_renderer.rs
```diff
@@ -54,6 +54,12 @@ impl Default for ConfigRenderer {
     fn default() -> Self {
         let mut template_engine = Handlebars::new();
         template_engine.set_strict_mode(true); // enable throwing render errors if context data is valid
+        template_engine.register_escape_fn(|s| s.into()); // prevent escaping like double quotes to &quot; ...
```
krucod3, line 57: Should we add a unit test for this too?

krucod3 (commented):
Looks good, just a couple of small things.

inf17101:
> Looks good, just a couple of small things.

Thank you for the review. I will fix the findings. I will add the new partial syntax to the user documentation.

server/src/ankaios_server/config_renderer.rs
inf17101: Done in [38d68a](https://github.com/eclipse-ankaios/ankaios/pull/445/commits/38d68af17fb217e8acfe11b299d925dbc90ee2d2)

krucod3:
Is this a standalone PR, or is it part of the current ticket #393? I was assuming the later, but if we want it as a standalone one, we should either add a description to it, or an create an issue and link it. (I would go for the description.)

inf17101:
> Is this a standalone PR, or is it part of the current ticket #393? I was assuming the later, but if we want it as a standalone one, we should either add a description to it, or an create an issue and link it. (I would go for the description.)

The issue was detected within #393 but the fix is independent. I will add a description.

doc/docs/usage/manifest/config-rendering.md
```diff
@@ -0,0 +1,112 @@
...
+events {
+    worker_connections  1024;
+}
+
+http {
+    server {
+...
+```
+
+By using the `indent` control structure, incorrect indentation will be prevented and the line indentation of the current context will be considered. This results in an error-free YAML file.
```
krucod3, line 91: Let's stay positive: 
```suggestion
By using the `indent` control structure, the line indentation of the current context will be considered which results in an error-free YAML file.
```

krucod3 (commented):
Reviewed, small findings documented. 

There are also some value duplications in the new test for the escaped string, but it's ok for a test.

doc/docs/usage/manifest/config-rendering.md
```diff
@@ -0,0 +1,112 @@
+# Shareable configuration
+
+Ankaios supports the shareable configuration approach, which allows configurations to be defined once and assigned to any number of workloads. It uses the [handlebars](https://github.com/sunng87/handlebars-rust) templating syntax to extend certain fields of workloads. For a detailed look at supported workload fields and a basic example of sharing configuration with workloads, see [here](../../reference/startup-configuration.md).
```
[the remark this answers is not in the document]
inf17101: Ok, yes my first approach avoided duplication and linked to the startup-configuration instead. But I think it is good to mention it explicitly here again. It increases usability and you find again all in one place.

inf17101:
> Reviewed, small findings documented.
> 
> There are also some value duplications in the new test for the escaped string, but it's ok for a test.

I will make constants but not global, because it is only needed inside the escape test.

inf17101:
> > Reviewed, small findings documented.
> > There are also some value duplications in the new test for the escaped string, but it's ok for a test.
> 
> I will make constants but not global, because it is only needed inside the escape test.

Done.

krucod3 (approved):
LGTM 👍

Issue #446, opened 2025-01-31
Workload can't be deleted


sboett-dev:
When the Ankaios server machine is restarted and shortly afterwards a workload shall be shifted via control interface to another machine, the original workload deletion fails.

## Current Behavior
After the machine reboots the workload shall be shifted from one machine to another: it is started on the new machine but the deletion on the original machine fails. 

## Expected Behavior
After the machine reboots the workload shall be shifted from one machine to another: it is started on the new machine and successfully deleted on the original machine. 

## Steps to Reproduce
1. create 3 machines: one with the Ankaios server and an agent (node A), 2 machines with an agent only (node B and C)
2. on node A: have an Ankaios orchestrated containerized composer application (started with podman-kube) that can create/update workloads over the control interface when it is started. Make the composer application start a workload on node B after its start. Whenever the application is restarts, make it start/update the very same workload on node C (workload shift).
3. reboot the node A machine (which runs the Ankaios server and control interface composer application), let the 2 oder machines run
4. after reboot, the composer application shifts the workload from node B to node C automatically (as defined in 2.)
5.  watch it create the workload successfully on node C but fail to delete the workload on node B

## Context (Environment)
- happened on AWS EC2 instances with EWAOL, others not tested

## Logs
I submitted the (very long) logs to @krucod3 already, here are some excerpts that match the description above:
```
Creating 'podman-kube' workload 'test_56dddd75ef6331e9'.

Agent 'nodeC' received UpdateWorkloadState: UpdateWorkloadState { workload_states: [WorkloadState { instance_name: WorkloadInstanceName { agent_name: "nodeB", workload_name: "test_56dddd75ef6331e9", id: "cbf4e956764028de377b0cab3169fade2b6aea9f496849d615d2f3a7d4736931" }, execution_state: ExecutionState { state: Stopping(RequestedAtRuntime), additional_info: "" } }] }

The server reports workload state 'ExecutionState { state: Stopping(RequestedAtRuntime), additional_info: "" }' for the workload 'test_56dddd75ef6331e9' in the agent 'nodeB'

The server reports workload state 'ExecutionState { state: Stopping(DeleteFailed), additional_info: "Error: no pod with name or ID test_56dddd75ef6331e9-pod-0 found: no such pod. Execution of 'podman \"kube\" \"down\" \"-\"'" }' for the workload 'test_56dddd75ef6331e9' in the agent 'nodeB'

[...]

PR #448, opened 2025-02-04
Fix the handling of unneeded workloads at server restart

krucod3:
Issues: #446

We cannot properly handle unneeded workloads in a restart situation because we could have missed information during the downtime. The delete dependencies are not available in such scenarios and the only thing we can do is to delete the workload without waiting.

Still, the system **shall not be left in a dirty state** if we can avoid this, so in case of a server restart the existing workload shall be properly deleted using the workload object. This ensures that:
* Control Interface pipes are deleted
* Config files for the mounts are deleted
* Workload state checker is deleted
* Workload object is not ghosting in the agent

# Definition of Done

### DIFF
--- a/agent/src/runtime_manager.rs
+++ b/agent/src/runtime_manager.rs
@@ -313,8 +313,7 @@ impl RuntimeManager {
                             }
                         } else {
                             let workload_name = workload_state.instance_name.workload_name();
-                            log::info!("Existing workload '{}' is not needed.", workload_name);
-                            // No added workload matches the found running one => delete it
+                            log::info!("Found existing workload '{}' is not needed.", workload_name);
                             // [impl->swdd~agent-existing-workloads-delete-unneeded~2]
                             if let Some(workload) = self.workloads.remove(workload_name) {
                                 if let Err(err) = workload.delete().await {
--- a/agent/src/runtime_manager.rs
+++ b/agent/src/runtime_manager.rs
@@ -312,16 +312,28 @@ impl RuntimeManager {
                                     .push(ReusableWorkloadSpec::new(new_workload_spec, None));
                             }
                         } else {
+                            let workload_name = workload_state.instance_name.workload_name();
+                            log::info!("Existing workload '{}' is not needed.", workload_name);
                             // No added workload matches the found running one => delete it
-                            // [impl->swdd~agent-existing-workloads-delete-unneeded~1]
-
-                            // workload states are allowed to send because the workload is not created anymore afterwards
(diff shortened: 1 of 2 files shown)


agent/src/runtime_manager.rs
```diff
@@ -312,16 +312,28 @@ impl RuntimeManager {
                                     .push(ReusableWorkloadSpec::new(new_workload_spec, None));
                             }
                         } else {
+                            let workload_name = workload_state.instance_name.workload_name();
+                            log::info!("Existing workload '{}' is not needed.", workload_name);
                             // No added workload matches the found running one => delete it
```
krucod3, line 317:
```suggestion
                            let workload_name = workload_state.instance_name.workload_name();
                            log::info!("Found existing workload '{}' is not needed.", workload_name);
```

krucod3, line 317:
```suggestion
                            let workload_name = workload_state.instance_name.workload_name();
                            log::info!("Found existing workload '{}' is not needed.", workload_name);
```

agent/src/runtime_manager.rs
```diff
@@ -312,16 +312,27 @@ impl RuntimeManager {
...
-                            runtime.delete_workload(
-                                workload_state.instance_name,
-                                &self.update_state_tx,
-                                REPORT_WORKLOAD_STATES_FOR_WORKLOAD,
-                            );
+                            let workload_name = workload_state.instance_name.workload_name();
+                            log::info!("Found existing workload '{}' is not needed.", workload_name);
+                            // [impl->swdd~agent-existing-workloads-delete-unneeded~2]
+                            if let Some(workload) = self.workloads.remove(workload_name) {
+                                if let Err(err) = workload.delete().await {
```
inf17101, line 319: Optionally, the `unwrap_or_else` might help to hide the additional nested if here for logging the error.

inf17101 (commented):
Review finished. One optional improvement suggestion. You can just resolve the conversation if you like the explicit 'if' more for logging the error.

agent/src/runtime_manager.rs
krucod3: I have to admit, I took the block from a different place in the same file and did not though that much about it.
It can be changed to a `unwrap_or_else`, but the code is exactly as long and as readable as before. In my eyes this is just a syntactic preference of the writer.
To spare additional effort, I would leave it as is if you don't mind.

inf17101 (approved):
LGTM 👍

Issue #228, opened 2024-04-04
ank get workloads shall support a watch option


krucod3:
## Description

Currently `ank get workloads` does a single request of the state and presents it to the user. A watch is possible with the Linux watch command.
We can add a flag to the `get workloads` command specifying that the CLI shall watch for updates.

## Goals

Improve the CLI usability.

## Final result

The implementation can be done smartly by reacting to update workload states instead of polling for the complete state all the time.

We can get the state at the beginning and then listen for UpdateWorkloadState messages. 
Once a message for an unknown to the CLI workload comes, the CLI can request the compete state again and update the table.

With this implementation the CLI will shall all state changes and be independent of a sampling rate

### Summary

**To be filled when the final solution is sketched.**

### Tasks

- [ ] Task 1
- [ ] Task 2
- [ ] ...

krucod3:
The watch can be forever or until an end state (running, succeeded or failed).

krucod3:
The options for the watch could be `-w` and `--watch`.

MahmoudMohsenEgy:
I would like to work on this issue. Assign it to me if possible. Thanks!

inf17101:
The PR was merged to main. Issue can be closed. Thanks a lot for the great effort and work of @MahmoudMohsenEgy.

PR #463, opened 2025-03-05
Add watch option to 'ank get workloads' command

MahmoudMohsenEgy:
Issues: #228

This PR introduces a watch option to the `ank get workloads` command, allowing the CLI to continuously fetch and update the workload state instead of fetching it once and exiting.

Changes: 

- Listens for `UpdateWorkloadState` messages via Ankaios server's
- Dynamically updates the CLI table when changes occur
- Uses `BTreeMap` for automatic sorting and efficient O(log n) lookup
- Maintains initial filters (agent, state, workload name) during updates

Usage:

```
ank get workloads --watch
```
```
ank get workloads -w
```
```
ank get workloads --agent agent_A --state 'running(ok)' --watch
```

# Definition of Done


inf17101:
@MahmoudMohsenEgy: Thanks a lot for your great contribution. We are currently reviewing your PR. From the first view point of view, the unit tests could be enhanced with testing the new functionality. We have also [contribution guidelines](https://eclipse-ankaios.github.io/ankaios/main/development/rust-coding-guidelines/) you might take a look or have already. If you have any question or if you miss something feel free to ask. We are happy to help.

ank/src/cli_commands/get_workloads.rs
```diff
@@ -56,18 +59,89 @@ impl CliCommands {
...
-        let table_rows: Vec<WorkloadTableRow> = workload_infos.into_iter().map(|x| x.1).collect();
+        let mut workloads_table_data: BTreeMap<String, WorkloadTableRow> = workload_infos
+        .into_iter()
+        .map(|(i_name, row)| (i_name.to_string(), row))
+        .collect();
+
+
+        if !watch {
+            // // [impl->swdd~cli-shall-present-list-of-workloads~1]
+            let table_rows: Vec<WorkloadTableRow> = workloads_table_data.values().cloned().collect();
```
inf17101, line 70: The clone is actually only needed because the `workloads_table_data` is used after the if-condition checking the watch. By refactoring the watch code into separate function and defining an if-else you can move the data in instead of cloning it.

ank/src/cli_commands/get_workloads.rs
```diff
@@ -11,33 +11,36 @@
...
 
 use super::cli_table::CliTable;
 use super::workload_table_row::WorkloadTableRow;
 use super::CliCommands;
 
+use std::collections::BTreeMap;
+
 impl CliCommands {
     // [impl->swdd~cli-provides-list-of-workloads~1]
     pub async fn get_workloads_table(
```
inf17101, line 24: This method has now some side-effects when putting the watch implementation into it in addition. Because, if the watch is enabled it never returns and does not reflect that it is a getter `get_...` anymore and if the watch is disabled it returns a table output string somehow.

My suggestion is to keep the existing `get_workloads_table` as it was and define a new method doing the watch logic. In this way you can get rid of the side-effect and the `main.rs` (the calling side) can handle the check for `watch` option and dispatch to the corresponding function. In this way you can also get rid of the 4th argument.

What do you think?

ank/src/cli_commands/get_workloads.rs
```diff
@@ -56,18 +59,89 @@ impl CliCommands {
...
+                                updated_workloads_btree.retain(|_k, row| row.agent == *agent_name);
+                            }
+
+                            if let Some(state) = &state {
+                                updated_workloads_btree.retain(|_k, row| row.execution_state.eq_ignore_ascii_case(state));
+                            }
+
+                            if !workload_name.is_empty() {
+                                updated_workloads_btree.retain(|_k, row| workload_name.iter().any(|wn| wn == &row.name));
+                            }
```
inf17101, line 112: One suggestion to avoid code duplication with existing code. We have already a filtering code for workload table entries in the `get_workloads_table` method. My suggestion is to extract the fetch of the workloads plus the post-filtering and sorting into a new method that is called something like `get_filtered_workloads` and just use it in the two places (first place inside the `get_workloads_table` and the second usage in the `watch` code.

ank/src/cli_commands/get_workloads.rs
```diff
@@ -56,18 +59,89 @@ impl CliCommands {
...
+            return Ok(table);
+        }
+
+        update_table(&workloads_table_data);
+        // [impl->swdd~cli-watches-workloads~1]
+        loop {
+            let workload_update_result = self.server_connection.read_next_update_workload_state().await;
+
+            match workload_update_result {
+                Ok(update) => {
```
inf17101, line 88: The `Ok` match branch looks deeply nested. My suggestion is to extract the whole `Ok` part (besides `update_table(&workloads_table_data);`) into a method called something like `async fn update_workload_table_entries(...)` doing the continuous updates of the workload table data.

ank/src/cli_commands/get_workloads.rs
```diff
@@ -56,18 +59,89 @@ impl CliCommands {
...
-        // [impl->swdd~cli-shall-present-list-of-workloads~1]
-        let table_rows: Vec<WorkloadTableRow> = workload_infos.into_iter().map(|x| x.1).collect();
+        let mut workloads_table_data: BTreeMap<String, WorkloadTableRow> = workload_infos
+        .into_iter()
+        .map(|(i_name, row)| (i_name.to_string(), row))
+        .collect();
+
+
+        if !watch {
+            // // [impl->swdd~cli-shall-present-list-of-workloads~1]
```
inf17101, line 69: One comment string is enough.
```suggestion
            // [impl->swdd~cli-shall-present-list-of-workloads~1]
```

ank/src/cli_commands/get_workloads.rs
```diff
@@ -56,18 +59,89 @@ impl CliCommands {
...
+            let table_rows: Vec<WorkloadTableRow> = workloads_table_data.values().cloned().collect();
+
+            // // [impl->swdd~cli-shall-present-workloads-as-table~1]
```
inf17101, line 72:
```suggestion
            // [impl->swdd~cli-shall-present-workloads-as-table~1]
```

inf17101 (commented):
Review done.
Thank you for your great contribution. I have tested it and it works really good. I have put some small suggestions to improve the integration of the new feature into existing code structures.

In addition, since Ankaios is targeted for automotive, we believe that requirements tracing is needed. I have seen you have linked existing requirements which is really good (Thank you for that). In addition, we would need one additional swdd requirement in the [ank get workloads swdd section](https://github.com/eclipse-ankaios/ankaios/blob/main/ank/doc/swdesign/README.md#ank-get-workload) telling that the ank-cli has the new `watch` feature and to call the watch functionality in the code at the end. You can simply enhance this section, add a new swdd requirement there and link it with `// impl->` to the function call in `main.rs`. Could you create the new swdd? The uml diagram might also need a small udpate. 

If you need further support or you have questions, you are welcome.

Thank you for your good work 👍

ank/src/cli_commands/get_workloads.rs
MahmoudMohsenEgy: I agree with the idea of reducing duplication. The challenge is that we’re handling two different data structures: one is a `Vec` of tuples in `get_workloads_table` and the other is a `BTreeMap` in the `watch` code. A single get_filtered_workloads would need to account for these differences, complicating things.

Instead, I propose extracting the filtering logic into a helper method `is_workload_valid` that accepts the agent, state, and workload name as parameters and returns a boolean. This method can then be passed to the `retain` calls in both contexts, without forcing a one-size-fits-all approach.

In the first part it will be 

```
  workload_infos.get_mut().retain(|wi| is_workload_valid(&wi.1, &agent_name, &state, &workload_name));
```

in the second part it will be 
```
updated_workloads_btree.retain(|_k, row| is_workload_valid(row, &agent_name, &state, &workload_name));
```

What do you think?

ank/src/cli_commands/get_workloads.rs
```diff
@@ -56,18 +59,89 @@ impl CliCommands {
...
-            .table_with_wrapped_column_to_remaining_terminal_width(
-                WorkloadTableRow::ADDITIONAL_INFO_POS,
-            )
-            .unwrap_or_else(|_err| CliTable::new(&table_rows).create_default_table()))
+        }
     }
 }
 
+fn update_table(table_data: &BTreeMap<String, WorkloadTableRow>) {
+    let rows: Vec<_> = table_data.values().cloned().collect();
```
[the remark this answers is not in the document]
MahmoudMohsenEgy: Yes, you're right. I mistakenly cloned the data here. Thank you for alerting me to this.

ank/src/cli_commands/get_workloads.rs
inf17101: Resolved in [b4d3ddc](https://github.com/eclipse-ankaios/ankaios/pull/463/commits/b4d3ddc21153780c6fd550eb20aa5408350afe06)

ank/src/cli_commands/get_workloads.rs
inf17101: Resolved in [b4d3ddc](https://github.com/eclipse-ankaios/ankaios/pull/463/commits/b4d3ddc21153780c6fd550eb20aa5408350afe06)

ank/src/cli_commands/get_workloads.rs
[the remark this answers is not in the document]
inf17101: Resolved in [7bbc151](https://github.com/eclipse-ankaios/ankaios/pull/463/commits/7bbc151fc656d701784c18f2e553f620f865c8a1)

ank/src/cli_commands/get_workloads.rs
MahmoudMohsenEgy: Thank you for the suggestion. As I was thinking through the implementation of the --watch option, I believe I’ve found a specific scenario that requires careful handling.

In the `--watch` Flow, There Are Two Distinct Cases:

1. New Workloads:
When a new workload is detected, the code fetches the complete state again, applies the filters, and transforms the data into a `BTreeMap`. This ensures that new workloads are properly filtered before being added to the table.

2. Existing Workload Updates:
When an existing workload’s state changes (e.g., from `Pending` to `Running`), the workloads are already stored in the `BTreeMap`. At this point, we need to **reapply the filters directly** to the `BTreeMap` to ensure the updated workload still meets the filtering criteria.


Example Scenario

Consider the following command:

`ank get workloads --state "Pending(WaitingToStart)" --watch`

the initial table might look like this:

| WORKLOAD | NAME   | AGENT   | RUNTIME                 | EXECUTION STATE          |
|----------|--------|---------|-------------------------|--------------------------|
| workload1   | agent_A | podman  | Pending (WaitingToStart)           |                          |
| workload2   | agent_A | podman  | Pending (WaitingToStart) |                          |
| workload3   | agent_A | podman  | Pending (WaitingToStart)         |                          |


If `workload2` state is changed to  `Running(Ok)`, it should no longer appear in the table because it no longer matches the `--state "Pending(WaitingToStart)"` filter. However, the current implementation does not reapply the filters to existing workloads during updates, which could lead to inconsistencies as in existing workload updates workloads are already in `BTreeMap` so filtering it should occur be in-place.

So in my suggestion separating data structures from the filtering logic so that it can be invoked in the retain method will be a better approach.

I've committed this change, could you review it and let me know If you notice anything in this case please tell me.

inf17101: I like the solution in [c9dab16](https://github.com/eclipse-ankaios/ankaios/pull/463/commits/c9dab16fb57cb1fecc0111af87c07eefa02550df).

ank/src/cli_commands/get_workloads.rs
inf17101: I can see that the clone is removed now, but putting the new watch functionality into a separate function is still open.

ank/src/cli_commands/get_workloads.rs
```diff
@@ -11,63 +11,145 @@
...
+        }
+    }
+
+    if !workload_names.is_empty() && !workload_names.iter().any(|wn| wn == &row.name){
+            return false;
+
+    }
+
+    true
+}
```
inf17101, line 151: Now, that we have centralized the filtering. Can you maybe reverse the checks of the existing code so that the if branches return `true` and at the end of the function it returns `false` ? This would be easier to read.

inf17101: No need to change reverse the checks, otherwise the retain method has to be called multiple times.

ank/src/cli_commands/get_workloads.rs
```diff
@@ -113,6 +86,46 @@ impl CliCommands {
...
         }
     }
+
+    async fn process_workload_updates(
+        &mut self,
+        update: UpdateWorkloadState,
+        agent_name: &Option<String>,
+        state: &Option<String>,
+        workload_name: &[String],
+        workloads_table_data: &mut BTreeMap<String, WorkloadTableRow>,
```
inf17101, line 96: Instead of an in-out variable, can you just return it and assign it on the caller side?

ank/src/cli_commands/get_workloads.rs
inf17101: Author has fixed it in [8b879aa](https://github.com/eclipse-ankaios/ankaios/pull/463/commits/8b879aadb124742ec3efd9e00fc82a2874edbd31)

ank/src/cli_commands/get_workloads.rs
MahmoudMohsenEgy: I changed it in [41e931c](https://github.com/eclipse-ankaios/ankaios/pull/463/commits/41e931cbab5397585f5521b26e775c9e6574d9af)

inf17101: Reviewed, looks good.

ank/src/cli_commands/get_workloads.rs
MahmoudMohsenEgy: I appreciate your detailed review!

How I handled this: 

- I split the functionality into two distinct methods:

    `get_workloads_table`: Pure getter that immediately returns the table data as it was.
    `watch_workloads`: Dedicated method handling continuous watching.
This maintains the original contract of `get_workloads_table` while isolating watch-specific logic.

- Created a helper method `fetch_filtered_workloads` that handles common fetching/filtering logic

- Kept related methods in `get_workloads.rs` since watch functionality is inherently tied to `ank get workloads` command and represents a variant of the same core operation.
 
- The calling side`main.rs` now explicitly handles the checks and chooses between the methods.

Thanks for your thoughtful input!

MahmoudMohsenEgy: I appreciate your detailed review!

How I handled this: 

- I split the functionality into two distinct methods:

    `get_workloads_table`: Pure getter that immediately returns the table data as it was.
    `watch_workloads`: Dedicated method handling continuous watching.
    
    This maintains the original form of `get_workloads_table` while isolating watch logic.

- Created a helper method `fetch_filtered_workloads` that handles common fetching/filtering logic

- Kept related methods in `get_workloads.rs` since watch functionality is inherently tied to `ank get workloads` command and represents a variant of the same core operation.
 
- The calling side`main.rs` now explicitly handles the checks and chooses between the methods.

Thanks for your thoughtful input!

inf17101: Great work, thank you. Resolved in [6cc98f3](https://github.com/eclipse-ankaios/ankaios/pull/463/commits/6cc98f399078b69bc4b972f6e2e2f5699be35f01)

ank/src/cli_commands/get_workloads.rs
inf17101: Resolved in [6cc98f3](https://github.com/eclipse-ankaios/ankaios/pull/463/commits/6cc98f399078b69bc4b972f6e2e2f5699be35f01)

ank/src/cli_commands/get_workloads.rs
```diff
@@ -25,49 +28,150 @@ impl CliCommands {
...
+        Ok(CliTable::new(&table_rows)
+            .table_with_wrapped_column_to_remaining_terminal_width(
+                WorkloadTableRow::ADDITIONAL_INFO_POS,
+            )
+            .unwrap_or_else(|_err| CliTable::new(&table_rows).create_default_table()))
+
+
+    }
+
+    // [impl->swdd~cli-watches-workloads~1]
```
inf17101, line 46: Unfortunately, this existing swdd cannot be linked to the new `watch` functionality, because the last item does not match. The new `watch` functionality of the `get workloads` command watches infinitely until a user terminates the watch command. The existing swdd covers more the watch of the workload updates (delete, apply, set state, run) and the watch will stop when the workloads have reached a desired state.

Proposal:

1. Remove the linkage of `swdd~cli-watches-workloads~1` from the new get workloads watch logic
2. Create a new swdd `swdd~cli-get-workloads-with-watch~1`. The precondition can be something like `When the CLI requests a list of workload with the `watch` option enabled, the CLI shall: ...` followed by a bullet point list of the actions your new code when doing the infinite `watch`.
3. Rename the existing swdd from `swdd~cli-watches-workloads~1` to `swdd~cli-watches-workloads-on-updates~1` (search & replace with editor) to make the differences between those two swdds more clear, since we do not re-use the code between the different watch logic.
4. Link the new swdd `swdd~cli-watches-workloads-on-updates~1` to this function and to the caller (main.rs)

MahmoudMohsenEgy: Thanks for the clarification! I’ve finished the changes required:
- Removed the linkage of `swdd~cli-watches-workloads~1` from the `ank get workloads --watch` logic.
- Created `swdd~cli-get-workloads-with-watch~1` under the `ank get workloads` section, describing the infinite watch behavior. You can check it [CLI shall provides continuous workload monitoring with watch mode](https://github.com/MahmoudMohsenEgy/ankaios/tree/228_watch_option_support_for_get_workloads/ank/doc/swdesign#cli-shall-provides-continuous-workload-monitoring-with-watch-mode).
- Renamed `swdd~cli-watches-workloads~1` to `swdd~cli-watches-workloads-on-updates~1`.
- Linked `swdd~cli-get-workloads-with-watch~1` to the relevant function in `get_workloads.rs`  and its caller in `main.rs`.

Let me know if anything else is needed!

ank/doc/swdesign/README.md
```diff
@@ -535,6 +535,27 @@ Needs:
...
+
+When the CLI requests a list of workload with the `watch` option enabled, the CLI shall continuously stream real-time workload updates until manually terminated by the user by following this steps:
+- Fetch and display the initial list of workloads from the Ankaios Server, applying user-defined filters (agent, state, workload names).
+- Continuously listen for `UpdateWorkloadState` messages and apply workload updates.
+- Dynamically update the table by:
+    - Adding newly created workloads.
+    - Removing workloads marked as `Removed`.
+    - Modifying existing workloads' execution states or metadata.
+- Optimize rendering to refresh only affected table rows, preventing full-screen reload.
+- Execute infinitely until explicitly terminated by the user using `Ctrl+C`.
```
inf17101, line 549: Thank you making the swdd changes.

I have a few further suggestions:
- the first action might be reduced a little bit to something like `..., the CLI shall continuously watch for workload changes by:`
- The part while listening to workload state updates (bullet point 2 and 3) are related to each other and the dynamic workload table updates underly some conditions. I suggest to split the swdd into two swdds to add the conditions properly. The second one separates the "process_workload_updates" logic while the first one says that it continuously listens to workload state updates to update the workload table data.
- the bullet point with `Optimize rendering to refresh only ...` can be removed
- the last bullet point with `Ctrl+C` we can put as comment
- the filtering logic can be also mentioned in a comment

Would you do the changes again? If you need support or detailed suggestions let me know.

ank/src/cli_commands/get_workloads.rs
inf17101: Thanks a lot for your efforts and for doing the swdd changes on top. I have reviewed the swdd and provided some suggestions.

ank/doc/swdesign/README.md
MahmoudMohsenEgy: Thanks for your comments! I appreciate the feedback and always happy to improve to achieve best practices.

ank/doc/swdesign/README.md
```diff
@@ -535,6 +535,29 @@ Needs:
...
+- Fetch and display the initial list of workloads from the Ankaios Server.
+- Continuously listen for `UpdateWorkloadState` messages.
+- Dynamically update the table by:
+    - Adding newly created workloads.
+    - Removing workloads marked as `Removed`.
+    - Modifying existing workloads' execution states or metadata.
+
+Comment:
+- User termination via `Ctrl+C`.
+- Supports user-defined filters (agent, state, names).
```
inf17101, line 551:
```suggestion
Comment:
The `watch` mode executes infinitely until the user explicitly terminates it using `Ctrl+C`.
All received workload updates are filtered with user-defined parameters (agent, state, workload names).
```

inf17101: Taken over in [34a0779](https://github.com/eclipse-ankaios/ankaios/pull/463/commits/34a077965feba681c4f45726d6f64e6cfe71a947)

ank/doc/swdesign/README.md
```diff
@@ -535,6 +535,45 @@ Needs:
...
+- CliCommands
+
+Needs:
+- impl
+- utest
+
+#### CLI processes workload state updates for watch mode
+`swdd~cli-process-workload-updates~1`
+
+When processing UpdateWorkloadState messages during watch mode, the CLI shall:
```
inf17101, line 561:
```suggestion
When processing `UpdateWorkloadState` messages during watch mode, for each workload state the CLI shall:
```

ank/doc/swdesign/README.md
```diff
@@ -535,6 +535,45 @@ Needs:
...
+
+- Add new workloads: Full state re-fetch on unknown instances.
+- Update existing workloads: Modify workload data then revalidate filters.
+- Remove workloads: Immediate deletion on "Removed" state detection.
+- Reapply filters after every update cycle.
```
inf17101, line 566:
```suggestion
* either update the current list of workloads with a new list of workloads from the Ankaios Server if the received workload state is from a newly added workload
* or delete workload from the current list if the received state is `Removed`
* or update the current state with new state inside the current list
```

ank/doc/swdesign/README.md
```diff
@@ -535,6 +535,45 @@ Needs:
...
+
+Comment: Filters for unknown instances are applied prior to transitioning to the BTreeMap to ensure that workloads not matching the filters are excluded from insertion.
```
inf17101, line 568:
```suggestion
Comment:
All received workload data is filtered with user-defined parameters (agent, state, workload names).
```

inf17101 (commented):
Thanks a lot for your swdd adaptions. I have suggested some changes to better fit the style of existing requirements.

Please have a look at the comments and let me know if you are fine with it.

ank/doc/swdesign/README.md
```diff
@@ -535,6 +535,44 @@ Needs:
...
+#### CLI initiates continuous workload monitoring
+`swdd~cli-get-workloads-with-watch-init~1`
+
+When the CLI requests a list of workloads with the `watch` option enabled, the CLI shall continuously watch for workload changes by:
+
+- requesting the initial list of workloads from the Ankaios Server
+- displaying the workloads in an alphabetically sorted table
+- updating the initial list of workloads by continuously listening for `UpdateWorkloadState` messages
+
+Comment: The `watch` mode executes infinitely until the user explicitly terminates it using `Ctrl+C`. All received workload data is filtered with user-defined parameters (agent, state, workload names).
```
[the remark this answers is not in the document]
inf17101: I commit this directly. After `comment` we need a newline.

inf17101:
> @inf17101 Everything is fine, and I've committed it all.

Great, thanks a lot. Then I think the first swdd must be linked in the code and last but not least a few new utests.

MahmoudMohsenEgy:
> Great, thanks a lot. Then I think the first swdd must be linked in the code and last but not least a few new utests.

Yes, no worries, I'm working on it right now.

ank/doc/swdesign/README.md
inf17101: The swdd requirements have been adapted in [10feffa](https://github.com/eclipse-ankaios/ankaios/pull/463/commits/10feffa6a04c224d115474f2b6b8aa41ca5e1f71) and further commits.

inf17101:
> > Great, thanks a lot. Then I think the first swdd must be linked in the code and last but not least a few new utests.
> 
> Yes, no worries, I'm working on it right now.

I have checked the commit with the swdd linkage. It looks good to me.

MahmoudMohsenEgy:
> @MahmoudMohsenEgy I you update your fork with with eclipse-ankaios:main, then the build failure should disappear.

Done. Thanks for noticing me!

ank/src/cli_commands/get_workloads.rs
```diff
@@ -342,4 +486,187 @@ mod tests {
...
+                new_wl.runtime.clone(),
+                ExecutionState::running().to_string(),
+                String::new(),
+            ),
+        );
+        assert_eq!(table, expected_table);
+    }
+
+    #[tokio::test]
+    async fn utest_watch_workloads_removal() {
```
inf17101, line 590: You can simplify the test by just testing the removal with the `process_workload_updates` method directly instead of testing over the more complicated test setup via the public watch method.

I can see in the coverage check, that the branch for updating a workload's execution state in `process_workload_updates` is missing. Could you also add an utest for that? Just test this case with calling `process_workload_updates` in the utest. 

As a last utest, maybe we can test the failing `fetch_filtered_workloads` within the `watch_workloads` method (the question mark operator) over the public method. Then we have a satisfying coverage and we can finish this PR.

That would be great.

What do you think?

inf17101 (commented):
Thank you for your great efforts. The first utest looks really great. We are only a few changes away from the merge of the PR. If you need some additional support, let me know. I have published suggestions.

ank/src/cli_commands/get_workloads.rs
```diff
@@ -141,12 +151,30 @@ impl CliCommands {
...
+    mockall::lazy_static,
+    std::sync::{Arc, Mutex},
+};
+
+
+
 fn update_table(table_data: &BTreeMap<String, WorkloadTableRow>) {
-    let rows: Vec<&WorkloadTableRow> = table_data.values().collect();
-    let table = CliTable::new(&rows)
+    #[cfg(not(test))]
```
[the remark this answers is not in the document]
inf17101: Fixed in  b786133

ank/src/cli_commands/get_workloads.rs
```diff
@@ -342,4 +486,187 @@ mod tests {
 
         assert_eq!(cmd_text.unwrap(), expected_table_output);
     }
+
+
+    #[tokio::test]
+    async fn utest_watch_workloads_addition() {
+        // [utest->swdd~cli-get-workloads-with-watch~1]
```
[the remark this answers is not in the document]
inf17101: Fixed in b786133

ank/src/cli_commands/get_workloads.rs
```diff
@@ -342,4 +482,311 @@ mod tests {
...
+     );
+     assert_eq!(final_table_data.len(), 1, "Table should have one workload after removal");
+     assert_eq!(
+        final_table_data,
+        expected_table_data,
+        "The final table data did not match the expected table data."
+    );
+
+
+    }
```
inf17101, line 701: I think we can avoid the filter checks again if they are already covered in the other tests. This should be enough:

```suggestion
    #[tokio::test]
    async fn utest_process_workload_updates_removal() {
        let wl1 = generate_test_workload_spec_with_param(
            "agent_A".to_string(),
            "name1".to_string(),
            "runtime".to_string(),
        );

        let key_wl1 = wl1.instance_name.to_string();

        let mut initial_table_data = BTreeMap::new();

        initial_table_data.insert(
            key_wl1.clone(),
            WorkloadTableRow::new(
                wl1.instance_name.workload_name().to_string(),
                wl1.instance_name.agent_name().to_string(),
                wl1.runtime.clone(),
                ExecutionState::running().to_string(),
                String::new(),
            ),
        );

        let removed_workload_state = objects::WorkloadState {
            instance_name: wl1.instance_name.clone(),
            execution_state: ExecutionState::removed(),
        };
        let update_event = UpdateWorkloadState {
            workload_states: vec![removed_workload_state],
        };

        let mock_server_connection = MockServerConnection::default();
        let mut cmd = CliCommands {
            _response_timeout_ms: RESPONSE_TIMEOUT_MS,
            no_wait: false,
            server_connection: mock_server_connection,
        };

        let result_table_data = cmd
            .process_workload_updates(
                update_event,
                &None,
                &None,
                &Vec::new(),
                initial_table_data.clone(),
            )
            .await;

        assert!(
            result_table_data.is_ok(),
            "process_workload_updates failed: {:?}",
            result_table_data.err()
        );
        let final_table_data = result_table_data.unwrap();

        assert!(
            final_table_data.is_empty(),
            "The final table data should be empty after removal."
        );
    }
```

inf17101 (commented):
Unit tests are looking great. Just one suggestion to simplify the `utest_process_workload_updates_removal` utest.

inf17101 (approved):
Review done. Feature can be merged.

Issue #302, opened 2024-06-21
Support config file


windsource:
## Description

Currently the number of command arguments for `ank-server`, `ank-agent` and `ank` is increasing. For better configuration management it would be good to also support a config file which is read by the commands. The priority for parameters should be

1. Command arguments
2. Environment variables
3. Config file 

The config file could also contain parameters which we do not want to pass as arguments, like max. retry period or similar. In any case the config file should be optional.

For the server we could support a config file at `/etc/ankaios/server.toml`. 

For the agent maybe we can pass the location of the config file as command argument.

For the CLI a config file could be located at `$HOME/.config/ankaios/ank.toml`. Here contexts might be useful such that the user can switch the context for different clusters.

We should choose a format that can be parsed easy and fast. Maybe TOML is suited for that. I have tried the `toml` crate and it seems to be quite fast.

## Goals

Support config files as an additional method to configure `ank-server` (prio 1), `ank-agent` (prio 2) and `ank` (prio 3).

### Summary

**To be filled when the final solution is sketched.**

### Tasks

- [ ] Task 1
- [ ] Task 2
- [ ] ...

HorjuRares:
I can tackle this one.

HorjuRares:
A first variant of the server config file would be:
```
# This is a server config file with multiline strings

title = "Server Config File v1"

[server]
startup_config = './server/resources/startup_config.yaml'
# For the eventual consistency
MAX_RETRIAL_PERIOD = "300"

[server.mTLS]
mTLS_enabled = "true"

# TODO:
# INSTEAD OF THIS
# SHOULDN'T I ACTUALLY IMPLEMENT THE CREATION OF CERTIFICATES AND KEYS AS SEEN IN:
# https://eclipse-ankaios.github.io/ankaios/latest/usage/mtls-setup/

ANK_CA_PEM = '''-----BEGIN CERTIFICATE-----
MIHkMIGXAhRrcu4Zx8sBdvQZ53KiUCAV2/DSzjAFBgMrZXAwFTETMBEGA1UEAwwK
YW5rYWlvcy1jYTAeFw0yNDEyMTIwODE3MjJaFw0yNTAxMTEwODE3MjJaMBUxEzAR
BgNVBAMMCmFua2Fpb3MtY2EwKjAFBgMrZXADIQBiiJvCIsEl4yf1Iz2e30Fd/wDT
[...]

PR #467, opened 2025-03-17
Support ank config file

HorjuRares:
Issues: #302

Add support for the ank config file. Also, it implements the basis for supporting contexts in the future.

# Definition of Done


ank/src/ank_config.rs
```diff
@@ -0,0 +1,538 @@
...
+    }
+
+    pub fn update_with_args(&mut self, args: &AnkCli) {
+        if let Some(response_timeout) = args.response_timeout_ms {
+            self.response_timeout = response_timeout;
+        }
+
+        self.verbose = args.verbose;
+        self.quiet = args.quiet;
+        self.no_wait = args.no_wait;
```
krucod3, line 256: As these variables here have defaults, they overwrite the values from the config file even if we did not set them the the command line (or env variable).

ank/src/ank_config.rs
```diff
@@ -0,0 +1,538 @@
...
+    pub version: String,
+    pub response_timeout: u64,
+    pub verbose: bool,
+    pub quiet: bool,
+    pub no_wait: bool,
+    pub server_url: String,
+    pub insecure: bool,
+    pub ca_pem: Option<String>,
+    pub crt_pem: Option<String>,
+    pub key_pem: Option<String>,
```
krucod3, line 87: I suppose we don't need these public.

ank/src/ank_config.rs
```diff
@@ -0,0 +1,538 @@
...
+
+struct AnkConfigVisitor {
+    table_key: &'static str,
+}
+
+impl<'de> Visitor<'de> for AnkConfigVisitor {
+    type Value = AnkConfig;
+
+    fn expecting(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
+        formatter.write_str("a TOML map for AnkConfig with merged nested table fields")
```
krucod3, line 144: Actually the output is the merged nested table fields.
The visitor expects ... `a TOML with profiles in table fields.`

ank/config/ank.conf
```diff
@@ -5,35 +5,35 @@
...
-# insecure = 'false'
+insecure = false
 
 # The path to the ca certificate pem file.
-# ca_pem = '/etc/ankaios/certs/ca.pem'
+ca_pem = '/etc/ankaios/certs/ca.pem'
 
 # The path to CLI certificate pem file.
-# crt_pem = '/home/ankaios/.config/ankaios/ank.pem'
+crt_pem = '${XDG_CONFIG_HOME:-$HOME/.config}/ankaios/ank.pem'
```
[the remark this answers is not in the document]
HorjuRares: So should I also add the facility to expand the environment variables, or change the path as it was before?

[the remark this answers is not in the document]
HorjuRares: If the expansion is preferred, I will also implement it for the server and agent config files.

server/src/cli.rs
```diff
@@ -37,9 +37,9 @@ pub struct Arguments {
     #[clap(required = false, short = 'a', long = "address")]
     /// The address, including the port, the server shall listen at [default: 127.0.0.1:25551].
     pub addr: Option<SocketAddr>,
-    #[clap(short = 'k', long = "insecure", action=ArgAction::SetTrue, default_value_t = false, env="ANKSERVER_INSECURE")]
+    #[clap(short = 'k', long = "insecure", action=ArgAction::Set, num_args=0, default_missing_value="true", env = "ANKSERVER_INSECURE")]
```
krucod3, line 40: This should be per default false.

HorjuRares: It is actually not the value of the flag, but an indication that if the value is missing it has to be defaulted.

ank/config/ank.conf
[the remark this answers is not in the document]
krucod3: Leave it as it was. This is a feature outside of the current scope. if somebody needs, they can log an issue.

server/src/cli.rs
HorjuRares: https://docs.rs/clap/latest/clap/struct.Arg.html#method.default_missing_value

krucod3: OK, good to know about this flag. What is the difference to "action=ArgAction::SetTrue"?
How about also adding the "default_value_t = false" back to make it clear that the default is false.

And what is the difference after the change actually? It seem semantically the same.

HorjuRares: That is the stuff actually. We cannot use `default_value_t=false` as it won't default to `None` but to `false`. The actual difference at this point is the fact that if we specify `-k` we get a value of `Some(true)` while the absence defaults to `None`.

HorjuRares: We can use comments to show the actual value is `false` adding any other thing would disrupt the correct functioning.

krucod3: OK, understood 👍 
Let's leave it as is.

ank/src/ank_config.rs
```diff
@@ -0,0 +1,600 @@
...
+        if let Some(key_pem_path) = &ank_config.key_pem {
+            let key_pem_content = read_pem_file(key_pem_path, false)
+                .map_err(|err| ConversionErrors::InvalidCertificate(err.to_string()))?;
+            ank_config.key_pem_content = Some(key_pem_content);
+        }
+
+        Ok(ank_config)
+    }
+
+    pub fn update_with_args(&mut self, args: &AnkCli) {
```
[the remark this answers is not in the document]
HorjuRares: I've updated it in order to include and test AnkConfig for server_url.

ank/config/ank.conf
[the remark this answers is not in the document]
krucod3: We did not revert here to the state as it was before the changes:
```suggestion
crt_pem = '/home/ankaios/.config/ankaios/ank.pem'
```

krucod3 (changes requested):
One findings still left.

PR #476, opened 2025-04-23
Update base dev container

krucod3:
Updates:
* new Rust version 1.86
* updated just to 1.40

# Definition of Done

The PR shall be merged only if all items mentioned in [CONTRIBUTING.md](https://github.com/eclipse-ankaios/ankaios/blob/main/CONTRIBUTING.md#how-to-contribute) have been followed. In case an item is not applicable as described, please provide a short explanation in the description.

- [x] All steps in [CONTRIBUTING.md](https://github.com/eclipse-ankaios/ankaios/blob/main/CONTRIBUTING.md#how-to-contribute) have been handled

PR #478, opened 2025-04-29
Update control_interface_tester

krucod3:
The Cargo.lock is not required as the executable is build as part of Ankaios and used the main lock file.

# Definition of Done


tests/resources/control_interface_tester/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM ghcr.io/eclipse-ankaios/devcontainer-base:0.10.1 as compile
+FROM ghcr.io/eclipse-ankaios/devcontainer-base:0.10.3 AS compile
```
windsource, line 1: That would be new dev container base image which does not exist yet but created with #476 I think.

krucod3: Exactly. As discussed, this is not a problem as the builds must be triggered manually. The current one was done with a local version of the new dev container.

PR #485, opened 2025-05-07
Provided source package including dependencies with each release

windsource:
This PR adds a source archive as an asset for every release which includes the vendored dependencies, that have been used for that specific release. So everyone has the chance to rebuild Ankaios based on the same source and the same dependencies for that release.

For non-release builds (PRs and branches) the dependencies are not vendored in order not to increase the build time.

The newly create source package can later on be used to build packages for Linux distributions like Debian, Ubuntu or RHEL which usually require an archive with all dependencies.

A test release is available, see https://github.com/windsource/ankaios/releases/tag/v0.4.2 and the asset [`ankaios-vendored-source-0.4.2.tar.gz`](https://github.com/windsource/ankaios/releases/download/v0.4.2/ankaios-vendored-source-0.4.2.tar.gz).

# Definition of Done

### DIFF
--- a/.github/actions/use-vendored-source/action.yml
+++ b/.github/actions/use-vendored-source/action.yml
@@ -1,0 +1,12 @@
+name: Download and extract vendored source
+description: Download and extract vendored source. Usually only called for release builds
+runs:
+  using: "composite"
+  steps:
+    - name: Get vendored source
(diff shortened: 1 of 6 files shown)


GabyUnalaq (approved):
LGTM 👍

Issue #90, opened 2023-11-15
Ankaios command to fetch the logs of a workload


inf17101:
## Description

If a workload fails due to an application error within the workload then the user must use the commands of the runtime to fetch the logs of a container.

Ankaios does currently not support to fetch the logs about a workload. It only outputs the execution state, but not a reason why the workload has failed.

## Goals

The Ankaios CLI shall support fetching the internal logs of workloads (similiar to `podman logs` command).

### Summary

**To be filled when the final solution is sketched.**

### Tasks

- [ ] Task 1
- [ ] Task 2
- [ ] ...

windsource:
I think fetching is usually not enough. Besides fetching I would also like to get a stream similar to `podman logs --follow`. I know that this is more complicated but when we start with this feature we should include the streaming in the concept.

windsource:
I have checked `docker`, `podman` and `nerdctl` and all of them provide `stdout` and `stderr` in the logs which is expected.

krucod3:
I'll start with the concept here.

krucod3:
**Initial design after a sync with @christoph-hamm.**

This is a basic deployment diagram showing how the components will be connected:
![Image](https://github.com/user-attachments/assets/6121e129-6c6f-44e6-bd88-43d3c4e90809)

Here is also the PR with the proto changes: #461

Here some additional thoughts:
* What happens when an agent gests disconnected? 
**If a requesting agent gets disconnected, we should stop the campaign and all log collectors connected to it
If a collecting agent gets disconnected, this should be reflected in the requesting agents behavior towards the requester (shall be informed maybe) as logs will be missing.**
* Should we add additional services to tonic for the logs (separate endpoint vs currently existing for requests)? *No, use current endpoints*
We should definitely not block normal execution because of vast amounts of logs. Best is to run the log collection and forwarding in separate green threads (tasks) and also take care that the services handling the normal execution do not get clogged because of the logs. Ankaios shall remain responsive in all cases. If we assume 100 logs per workload per second and 200 workloads, tonic should still be able to handle it. => The logs will be handled in the current request response schema.
* One request is one log collection campaign
Should we allow requesting the logs from multiple workloads at the same time? **Why not? We will just start multiple collectors that collect for the same request Id.**
One collector per workload will make the implementation easier as we need to send a log entry per line per workload.
* We have to also decide on the dataflow especially when logs are requested and collected on the same agent. We have multiple options here: 
[...]

PR #496, opened 2025-05-27
Add log request authorization

GabyUnalaq:
Issues: #90 

This PR adds changes to the authorizer for allowing and denying log requests.

# Definition of Done

### DIFF
--- a/agent/src/control_interface/authorizer.rs
+++ b/agent/src/control_interface/authorizer.rs
@@ -14,29 +14,29 @@
 
 mod path;
 mod path_pattern;
-mod rule;
+mod rules;
+
+use std::{sync::Arc, vec};
 
 use common::{
     commands::Request,
     objects::{AccessRightsRule, ControlInterfaceAccess, ReadWriteEnum},
+    std_extensions::UnreachableOption,
 };
-use path_pattern::{AllowPathPattern, DenyPathPattern, PathPattern, PathPatternMatcher};
-#[cfg(not(test))]
-use rule::Rule;
+use path_pattern::{AllowPathPattern, DenyPathPattern, PathPatternMatcher};
+use rules::{LogRule, StateRule};
 
 #[cfg(test)]
 use mockall::mock;
-#[cfg(test)]
-use test::MockRule as Rule;
 
 #[derive(Clone, Default, Debug, PartialEq)]
 pub struct Authorizer {
-    allow_write_state_rule: Vec<Rule<AllowPathPattern>>,
-    deny_write_state_rule: Vec<Rule<DenyPathPattern>>,
-    allow_read_state_rule: Vec<Rule<AllowPathPattern>>,
-    deny_read_state_rule: Vec<Rule<DenyPathPattern>>,
-    allow_read_write_state_rule: Vec<Rule<AllowPathPattern>>,
-    deny_read_write_state_rule: Vec<Rule<DenyPathPattern>>,
+    state_allow_write: Vec<Arc<StateRule<AllowPathPattern>>>,
+    state_allow_read: Vec<Arc<StateRule<AllowPathPattern>>>,
+    state_deny_write: Vec<Arc<StateRule<DenyPathPattern>>>,
+    state_deny_read: Vec<Arc<StateRule<DenyPathPattern>>>,
+    log_allow: Vec<LogRule>,
+    log_deny: Vec<LogRule>,
 }
 
 #[cfg(test)]
@@ -70,11 +70,9 @@ impl Authorizer {
                 // [impl->swdd~agent-authorizing-all-elements-of-filter-mask-allowed~1]
                 field_mask.iter().all(|path_string| {
                     let path = path_string.as_str().into();
-                    let allow_reason = if let (true, reason) =
-                        self.allow_read_state_rule.matches(&path)
-                    {
-                        reason
-                    } else if let (true, reason) = self.allow_read_write_state_rule.matches(&path) {
+
+                    // [impl->swdd~agent-authorizing-matching-allow-rules~1]
+                    let allow_reason = if let (true, reason) = self.state_allow_read.matches(&path) {
                         reason
                     } else {
                         log::info!(
@@ -82,15 +80,13 @@ impl Authorizer {
                             path_string,
                             request.request_id
                         );
+                        println!("AAAAAAAAAA");
                         return false;
                     };
 
-                    let deny_reason = if let (true, reason) =
-                        self.deny_read_state_rule.matches(&path)
-                    {
-                        reason
-                    } else if let (true, reason) = self.deny_read_write_state_rule.matches(&path) {
-                        reason
+                    // [impl->swdd~agent-authorizing-matching-deny-rules~1]
+                    let deny_reason = if let (true, reason) = self.state_deny_read.matches(&path) {
+                            reason
                     } else {
                         log::debug!(
                             "Allow field mask '{}' of request '{}' as '{}' is allowed",
@@ -98,16 +94,18 @@ impl Authorizer {
                             request.request_id,
                             allow_reason
                         );
+                        println!("BBBBBBBBBB");
                         return true;
                     };
 
                     log::info!(
-                        "Deny field mask '{}' of request '{}',also allowed by '{}', as denied by '{}'",
+                        "Deny field mask '{}' of request '{}', also allowed by '{}', as denied by '{}'",
                         path_string,
                         request.request_id,
                         allow_reason,
                         deny_reason
                     );
+                    println!("CCCCCCCCCC");
                     false
                 })
             }
@@ -121,11 +119,9 @@ impl Authorizer {
                 // [impl->swdd~agent-authorizing-all-elements-of-filter-mask-allowed~1]
                 update_mask.iter().all(|path_string| {
                     let path = path_string.as_str().into();
-                    let allow_reason = if let (true, reason) =
-                        self.allow_write_state_rule.matches(&path)
-                    {
-                        reason
-                    } else if let (true, reason) = self.allow_read_write_state_rule.matches(&path) {
+
+                    // [impl->swdd~agent-authorizing-matching-allow-rules~1]
+                    let allow_reason = if let (true, reason) = self.state_allow_write.matches(&path) {
                         reason
                     } else {
                         log::info!(
@@ -136,11 +132,8 @@ impl Authorizer {
                         return false;
                     };
 
-                    let deny_reason = if let (true, reason) =
-                        self.deny_write_state_rule.matches(&path)
-                    {
-                        reason
-                    } else if let (true, reason) = self.deny_read_write_state_rule.matches(&path) {
+                    // [impl->swdd~agent-authorizing-matching-deny-rules~1]
+                    let deny_reason = if let (true, reason) = self.state_deny_write.matches(&path) {
                         reason
                     } else {
                         log::debug!(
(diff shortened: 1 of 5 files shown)


agent/src/control_interface/authorizer.rs
```diff
@@ -162,63 +152,139 @@ impl Authorizer {
...
+                        ReadWriteEnum::ReadWrite => {
+                            state_deny_read.push(rule.clone());
+                            state_deny_write.push(rule);
+                        }
+                        ReadWriteEnum::Nothing => {}
+                    }
+                }
+                AccessRightsRule::LogRule(log_rule) => {
+                    log_deny.push(log_rule.workload_names.clone().into());
+                }
```
christoph-hamm: This is a repetition of the code above. Extract it into a functions, as in the previous version.

agent/src/control_interface/authorizer.rs
```diff
@@ -162,63 +152,139 @@ impl Authorizer {
...
+
+                log::info!(
+                    "Deny log request '{}' as '{}' is allowed, but also denied by '{}'",
+                    request.request_id,
+                    allow_reason.unwrap_or_unreachable(),
+                    deny_reason.unwrap_or_unreachable()
+                );
+                false
+            }
```
christoph-hamm: This implements the allow/deny  rules a third time. This should really be moved to an extra function. Maybe we could create a struct containing the allow/deny rules and implement the logic as method there. The authorize could contain three instance of this struct, for write state, for read state and for logs.

agent/src/control_interface/authorizer.rs
```diff
@@ -162,63 +152,139 @@ impl Authorizer {
...
+                }
+
+                let allow_reason = logs_request
+                    .workload_names
+                    .iter()
+                    .find(|instance_name| {
+                        self.log_allow
+                            .iter()
+                            .any(|allow_rule| allow_rule.matches(instance_name.workload_name()))
+                    })
```
christoph-hamm: This will allow a request, if only for one of the requested workloads a rule exists. Hence, in the moment requesting logs for only one workload is allowed, logs for all workloads can be requested by simply including the one allowed one.

You should switch to `all` here. We will not get any information why a request was allowed, but this will be hard for a request with many workloads anyhow.

agent/src/control_interface/authorizer.rs
```diff
@@ -234,89 +300,112 @@ impl From<&ControlInterfaceAccess> for Authorizer {
...
+                    let rule = Arc::new(StateRule::<AllowPathPattern>::create(
+                        masks.iter().map(|x| x.as_str().into()).collect(),
+                    ));
+                    authorizer.state_allow_read.push(rule.clone())
+                }
+                RuleType::StateDenyRead(masks) => {
+                    let rule = Arc::new(StateRule::<DenyPathPattern>::create(
+                        masks.iter().map(|x| x.as_str().into()).collect(),
+                    ));
+                    authorizer.state_deny_read.push(rule.clone())
```
christoph-hamm: Remove the useless clones here.

agent/src/control_interface/authorizer.rs
```diff
@@ -234,89 +300,112 @@ impl From<&ControlInterfaceAccess> for Authorizer {
...
+                    ));
+                    authorizer.state_allow_read.push(rule.clone());
+                    authorizer.state_allow_write.push(rule.clone());
+                }
+                RuleType::StateDenyReadWrite(masks) => {
+                    let rule = Arc::new(StateRule::<DenyPathPattern>::create(
+                        masks.iter().map(|x| x.as_str().into()).collect(),
+                    ));
+                    authorizer.state_deny_read.push(rule.clone());
+                    authorizer.state_deny_write.push(rule.clone());
```
christoph-hamm: For this two cases, one clone is sufficient.

agent/src/control_interface/authorizer/path_pattern.rs
```diff
@@ -43,42 +43,29 @@ fn match_rule_with_path(rule: &impl PathPattern, other: &Path) -> (bool, PathPat
     )
 }
 
-pub trait PathPattern {
```
christoph-hamm, line 46: Why did you rearrange this file? It was not perfect, the `match_rule_with_path` should have been somewhere below, but besides this the ordering was ok.

As stated in our coding guidelines we try the follow the rules of "Clean Code" rules from "uncle Bob". Call dependencies should point downward. The caller should be above the callee.

agent/src/control_interface/authorizer/path_pattern.rs
```diff
@@ -126,10 +114,32 @@ impl PathPattern for DenyPathPattern {
...
+            },
+        }
+    }
+}
+
+impl From<Vec<PathPatternSection>> for DenyPathPattern {
+    fn from(value: Vec<PathPatternSection>) -> Self {
+        Self { sections: value }
+    }
+}
```
christoph-hamm: This is only used in the unit tests. You can move this to the test module.

agent/src/control_interface/authorizer/path_pattern.rs
```diff
@@ -115,6 +97,12 @@ impl From<&str> for DenyPathPattern {
     }
 }
 
+impl From<Vec<PathPatternSection>> for AllowPathPattern {
+    fn from(value: Vec<PathPatternSection>) -> Self {
+        Self { sections: value }
+    }
+}
+
```
christoph-hamm: This is only used in the unit tests. You can move this to the test module.

agent/src/control_interface/authorizer.rs
```diff
@@ -466,85 +570,154 @@ mod test {
...
+                follow: false,
+                tail: -1,
+                since: None,
+                until: None,
+            }),
+        };
+
+        let authorizer = Authorizer::default();
+        assert!(!authorizer.authorize(&request));
+        let authorizer = create_authorizer(&[RuleType::LogAllow(vec![WORKLOAD_NAME.into()])]);
```
christoph-hamm, line 588: Do we also have wildcards? If, we should also test them.

agent/src/control_interface/authorizer.rs
```diff
@@ -466,85 +570,154 @@ mod test {
...
+            request_content: common::commands::RequestContent::LogsRequest(LogsRequest {
+                workload_names: vec![],
+                follow: false,
+                tail: -1,
+                since: None,
+                until: None,
+            }),
+        };
+
+        let authorizer = Authorizer::default();
```
christoph-hamm, line 568: If we have wildcards, check if this is still denied with a wildcard allow.

agent/src/control_interface/authorizer.rs
```diff
@@ -466,85 +570,154 @@ mod test {
...
+    }
+
+    #[test]
+    fn utest_log_cancel_request() {
+        let request = Request {
+            request_id: "".into(),
+            request_content: common::commands::RequestContent::LogsCancelRequest,
+        };
+
+        let authorizer = Authorizer::default();
```
christoph-hamm, line 647: Also test this with rules.

agent/src/control_interface/authorizer.rs
GabyUnalaq: I could extract the three implementations for the CompleteStateRequest(read), UpdateStateRequest(write) and LogsRequest(logs) to three methods inside the `Authorizer itself`. In this method, no new struct is added, the rules are saved in the `Authorizer` as before, and the `authorize` method is simplified.
What do you think?

agent/src/control_interface/authorizer/path_pattern.rs
GabyUnalaq: It was confusing, it didn't look like it had any order. After consulting the Clean Code rules you mentioned, I can rearrange this again, to follow the order: traits, structs, enums, trait impl, non-trait impl, free functions.

Before, there were free functions, traits, trait impl, struct, impl for that struct, trait impl for that struct, another struct and so on. 

Is it ok to rearrange as I mentioned in the first paragraph?

agent/src/control_interface/authorizer.rs
GabyUnalaq: I am not sure if wildcards are present for the LogRules. If we are to add wildcards, we need to have a schema on how they should be used.

GabyUnalaq: After the discussion, we decided on adding wildcards and how to do it. The latest commit adds them and tests them.

agent/src/control_interface/authorizer/path_pattern.rs
christoph-hamm: I would not order it according to type. I would start with the public interface (often this will be traits). The implementation I would put directly after the structs they are implemented on, as this belongs together. In most Rust code I read non-trait implementation are before trait implementations. Hence if you have a struct A and a struct B, both implement trait I:
  
    trait I {...};
    struct A {...};
    impl A{...};
    impl I for A {...};
    struct B {...};
    impl B{...};
    impl I for B {...};

agent/src/control_interface/authorizer.rs
```diff
@@ -162,7 +154,64 @@ impl Authorizer {
...
+                    return false;
+                }
+
+                let allow_reason = logs_request
+                    .workload_names
+                    .iter()
+                    .find(|instance_name| {
+                        self.log_allow
+                            .iter()
+                            .all(|allow_rule| allow_rule.matches(instance_name.workload_name()))
```
christoph-hamm: I was wrong. The `all` does not solve the problem here. The follow test should pass, but the last assertion is failing. This test is not exhaustive. It still misses multiple allow and deny rules. So we will need even more tests.


    #[test]
    fn utest_log_requests_operations2() {
        fn request(workloads: &[&str]) -> Request {
            Request {
                request_id: "".into(),
                request_content: common::commands::RequestContent::LogsRequest(LogsRequest {
                    workload_names: workloads
                        .iter()
                        .map(|name| WorkloadInstanceName::new("", *name, ""))
                        .collect(),
                    follow: false,
                    tail: -1,
                    since: None,
                    until: None,
                }),
            }
        }

        fn allow_deny_authorizer(allow: Option<&[&str]>, deny: Option<&[&str]>) -> Authorizer {
            let mut rules = Vec::new();
            if let Some(allow) = allow {
                rules.push(RuleType::LogAllow(
                    allow.iter().map(|x| (*x).into()).collect(),
                ));
            }
            if let Some(deny) = deny {
                rules.push(RuleType::LogDeny(
                    deny.iter().map(|x| (*x).into()).collect(),
                ));
            }

            create_authorizer(&rules)
        }

        assert!(
            allow_deny_authorizer(Some(&["w1", "w2", "w3"]), Some(&["w3", "w4", "w5"]))
                .authorize(&request(&["w1"]))
        );
        assert!(
            !allow_deny_authorizer(Some(&["w1", "w2", "w3"]), Some(&["w3", "w4", "w5"]))
                .authorize(&request(&["w3"])),
        );
        assert!(
            allow_deny_authorizer(Some(&["w1", "w2", "w3"]), Some(&["w3", "w4", "w5"]))
                .authorize(&request(&["w1", "w2"])),
        );
        assert!(
            !allow_deny_authorizer(Some(&["w1", "w2", "w3"]), Some(&["w3", "w4", "w5"]))
                .authorize(&request(&["w1", "w3"])),
        );
        assert!(
            !allow_deny_authorizer(Some(&["w1", "w2", "w3"]), Some(&["w3", "w4", "w5"]))
                .authorize(&request(&["w1", "w6"])),
        );
    }

GabyUnalaq: I added this as an additional test, modified it a bit and added some additional cases. I also changed the logic of the allow_rules, and now the test passes, so it should be fine.

agent/src/control_interface/authorizer/path_pattern.rs
GabyUnalaq: I rearranged the file, please check if it's ok now.

agent/src/control_interface/authorizer/rules.rs
```diff
@@ -0,0 +1,232 @@
...
+            patterns: LogRule::validate_patterns(patterns),
+        }
+    }
+
+    pub fn matches(&self, workload_name: &str) -> bool {
+        for pattern in &self.patterns {
+            if pattern == workload_name || pattern == WILDCARD_SYMBOL {
+                return true;
+            } else if pattern.contains(WILDCARD_SYMBOL) {
+                let wildcard_pos = pattern.find(WILDCARD_SYMBOL).unwrap_or_unreachable();
```
christoph-hamm, line 66: Create a special `Pattern` type and move this logic there. I also think the splitting should be done in advance.
E.g you could create a `Pattern` trait and implement it for two types, one type for pattern without wildcard and one for patterns with wildcard. The wildcard check can also be quite simple, as `.starts_with("")` and `.ends_with("")` always return true.

agent/src/control_interface/authorizer/rules.rs
```diff
@@ -0,0 +1,232 @@
...
+                    wildcard_pos == 0 || workload_name.starts_with(&pattern[..wildcard_pos]);
+                let suffix_matches = wildcard_pos + WILDCARD_SYMBOL.len() == pattern.len()
+                    || workload_name.ends_with(&pattern[wildcard_pos + WILDCARD_SYMBOL.len()..]);
+                return prefix_matches && suffix_matches;
+            }
+        }
+        false
+    }
+
+    fn validate_patterns(patterns: Vec<String>) -> Vec<String> {
```
christoph-hamm, line 75: If you introduce a special `Pattern` type, you could instead us the conversion function. Just use `filter_map` or `flat_map` with the conversion function, and you should get the same result.

agent/src/control_interface/authorizer/rules.rs
GabyUnalaq: After a discussion, we decided to not add a new trait, but instead to refactor the code a bit.

christoph-hamm (approved):

PR #509, opened 2025-07-01
Implement logs cancel request upon agent gone

inf17101:
Issues: #90

When an agent notifies the server with an `AgentGone` message, for all workloads of the agent an `LogsCancelRequest` shall be sent to all other agents from the Ankaios Server. A data structure in the server tracks the mapping of the agents and request ids.

# Definition of Done

### DIFF
--- a/server/src/ankaios_server.rs
+++ b/server/src/ankaios_server.rs
@@ -17,7 +17,7 @@ mod cycle_check;
 mod delete_graph;
 mod server_state;
 
-use api::ank_base;
+use api::ank_base::{self};
 use common::commands::{Request, UpdateWorkload};
 use common::from_server_interface::{FromServerReceiver, FromServerSender};
 use common::objects::{
@@ -47,6 +47,8 @@ pub fn create_from_server_channel(capacity: usize) -> FromServerChannel {
     channel::<FromServer>(capacity)
 }
 
+type AgentLogRequestIdMap = std::collections::HashMap<String, std::collections::HashSet<String>>;
+
 pub struct AnkaiosServer {
     // [impl->swdd~server-uses-async-channels~1]
     receiver: ToServerReceiver,
@@ -54,6 +56,7 @@ pub struct AnkaiosServer {
     to_agents: FromServerSender,
     server_state: ServerState,
     workload_states_map: WorkloadStatesMap,
+    agent_log_request_id_map: AgentLogRequestIdMap,
 }
 
 impl AnkaiosServer {
@@ -63,6 +66,7 @@ impl AnkaiosServer {
             to_agents,
             server_state: ServerState::default(),
             workload_states_map: WorkloadStatesMap::default(),
+            agent_log_request_id_map: AgentLogRequestIdMap::default(),
         }
     }
 
@@ -178,6 +182,18 @@ impl AnkaiosServer {
                         )
                         .await
                         .unwrap_or_illegal_state();
+
+                    let request_ids_for_agent_gone = self
+                        .agent_log_request_id_map
+                        .remove(&agent_name)
+                        .unwrap_or_default();
+
+                    for request_id in request_ids_for_agent_gone {
+                        self.to_agents
+                            .logs_cancel_request(request_id)
+                            .await
+                            .unwrap_or_illegal_state();
+                    }
                 }
                 // [impl->swdd~server-provides-update-desired-state-interface~1]
                 ToServer::Request(Request {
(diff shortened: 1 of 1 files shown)


server/src/ankaios_server.rs
```diff
@@ -458,6 +550,7 @@ mod tests {
     const RUNTIME_NAME: &str = "runtime";
     const REQUEST_ID: &str = "request_1";
     const REQUEST_ID_A: &str = "agent_A@request_1";
+    const REQUEST_ID_B: &str = "agent_B@request_2";
```
krucod3, line 553: This should be `agent_A@request_2`. The comment here is only for completeness. As we are using the branch as a base for the next iteration, I have fixed the test with commit [8c6d216](https://github.com/eclipse-ankaios/ankaios/pull/520/commits/8c6d2162a17d09cd19d2225a5433595ffeaf0433).

server/src/ankaios_server.rs
```diff
@@ -1368,6 +1475,63 @@ mod tests {
...
+            create_from_server_channel(common::CHANNEL_CAPACITY);
+
+        let mut server = AnkaiosServer::new(server_receiver, to_agents);
+        server
+            .log_campaign_store
+            .expect_remove_agent_log_campaign_entry()
+            .once()
+            .return_const(Some(HashSet::from([
+                REQUEST_ID_A.to_owned(),
+                REQUEST_ID_B.to_owned(),
```
krucod3, line 1493: We are missing a `.with` here to check if the call was done with the correct agent name. Additionally as stated in the other comment we should have a second request id from the same agent and not from agent_B. As mentioned the comment is just for completeness. The fix is done with commit [8c6d216](https://github.com/eclipse-ankaios/ankaios/pull/520/commits/8c6d2162a17d09cd19d2225a5433595ffeaf0433) in the next iteration.

server/src/ankaios_server/log_campaign_store.rs
```diff
@@ -12,82 +12,212 @@
...
-                self.agent_log_request_ids_store.remove(&agent_or_cli_name);
-            } else {
-                entry.remove(request_id);
+        let request_id: RequestId = request_id.into();
+        log::debug!("Remove log campaign '{}'", request_id);
+
+        match request_id {
+            RequestId::CliRequestId(cli_request_id) => {
+                self.cli_log_request_id_store
+                    .remove(&cli_request_id.cli_name);
```
inf17101: Here the cli map does not support multiple request ids from the same cli-conn in case of a third party app overtakes the CLI part. And we need to proof the occurrence of the request id in the store itself like done for the agent and workload part.

server/src/ankaios_server/log_campaign_store.rs
```diff
@@ -106,17 +236,117 @@ mod tests {
...
+        }
+    }
+
     #[test]
-    fn utest_agent_log_connection_store_remove_all_logs_request_ids_for_agent() {
+    fn utest_cli_log_connection_store_insert_log_campaign() {
         let mut log_campaign_store = LogCampaignStore::default();
+
+        log_campaign_store.insert_log_campaign(CLI_REQUEST_ID_1.to_owned());
+        assert_eq!(log_campaign_store.cli_log_request_id_store.len(), 1);
```
inf17101, line 346: Would it be enough to assert the length only at the end to assert the total number of entries instead of after every insert? Maybe it fits also to the other insert utests.

inf17101:
> ` [b8dfa8a](/eclipse-ankaios/ankaios/pull/509/commits/b8dfa8a75d99f4d4d683f61b9e8ed686281a5509)`

Thanks for the review.
I have finished the review of your commits starting from [b8dfa8a](https://github.com/eclipse-ankaios/ankaios/pull/509/commits/b8dfa8a75d99f4d4d683f61b9e8ed686281a5509). I have left two comments.

server/src/ankaios_server/log_campaign_store.rs
krucod3: It is just a precautious measure to make sure the number is increasing as expected and no additional information is somehow added to the maps.
I would leave as is if you do not mind.

server/src/ankaios_server/log_campaign_store.rs
krucod3: Actually the whole design is done with that in mind. The cli_log_request_id_store does not have a HashSet of requests. 
The changes will spread through the whole implementation so I would suggest to make them in #520 in order to prevent conflicts.

server/src/ankaios_server/log_campaign_store.rs
inf17101: Ok, then let's keep it.

server/src/ankaios_server/log_campaign_store.rs
krucod3: Fixed in #520 with 9caed90

PR #511, opened 2025-07-03
Cancel log collection on broken collector

krucod3:
Issues: #90

# Definition of Done

### DIFF
--- a/agent/doc/swdesign/README.md
+++ b/agent/doc/swdesign/README.md
@@ -3215,10 +3215,10 @@ Needs:
 
 Status: approved
 
-The Ankaios Agent shall ensure, that Control Interface input pipes are opened and messages are read.
+The Ankaios agent shall ensure, that Control Interface input pipes are opened and messages are read.
 
 Comment:
-If the Ankaios Agent does not open and read the Control Interface input pipes, a Workload could block, trying to write the output pipe.
+If the Ankaios agent does not open and read the Control Interface input pipes, a Workload could block, trying to write the output pipe.
 
 Tags:
 - AgentManager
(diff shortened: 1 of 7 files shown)


agent/src/control_interface/output_pipe.rs
```diff
@@ -15,21 +15,40 @@
...
     file: Option<Sender>,
 }
 
+#[derive(Debug)]
+pub enum OutputPipeError {
+    ReceiverGone(io::Error),
+    Io(io::Error),
+}
+
+impl From<OutputPipeError> for io::Error {
```
christoph-hamm, line 43: Normally `From` implementations should be lossless. This one loses the information if it is a `ReceiverGone` or general `Io`. I think it is easy to accidentally lose the information using the `?` operator, as it performance in implicit conversion with `into`.

agent/src/control_interface/output_pipe.rs
```diff
@@ -38,19 +57,50 @@ impl OutputPipe {
...
     }
 
-    pub async fn write_all(&mut self, buf: &[u8]) -> io::Result<()> {
+    // [impl->swdd~agent-handles-control-interface-output-pipe-closed~1]
+    fn receiver_gone(err: &io::Error) -> bool {
+        // occurs when trying to write to a pipe that has no reader
+        err.kind() == ErrorKind::BrokenPipe
+        // occurs when trying to open the pipe for writing, but it is not open for reading
+        || err.raw_os_error() == Some(nix::libc::ENXIO)
+    }
```
christoph-hamm: This function should go below. (See vertical ordering in Clean Code)

agent/src/control_interface/output_pipe.rs
```diff
@@ -38,19 +57,50 @@ impl OutputPipe {
...
+
+    pub async fn write_all(&mut self, buf: &[u8]) -> Result<(), OutputPipeError> {
         if buf.is_empty() {
             return Ok(());
         }
+        let mut retries = 0;
```
christoph-hamm, line 65: I would use a for loop below. Makes it more obvious that this loop is limited.

agent/src/control_interface/output_pipe.rs
```diff
@@ -38,19 +57,50 @@ impl OutputPipe {
...
+                // [impl->swdd~agent-handles-control-interface-output-pipe-closed~1]
+                Err(err) if Self::receiver_gone(&err) => {
+                    if retries < CONTROL_INTERFACE_MAX_RETRIES {
+                        self.file = None;
+                        log::debug!("Broken pipe - the receiver is gone. Waiting for 'AGENT_RECONNECT_INTERVAL'ms before trying again.");
+                        sleep(Duration::from_millis(AGENT_RECONNECT_INTERVAL_MS)).await;
+                    } else {
+                        log::warn!("Failed to write to output pipe after multiple attempts");
+                        return Err(OutputPipeError::ReceiverGone(err));
+                    }
```
christoph-hamm, line 90: Make the else case on early return.

agent/src/control_interface/output_pipe.rs
```diff
@@ -150,4 +225,87 @@ mod tests {
...
+        });
+
+        {
+            let mut reading_side = super::OpenOptions::new().open_receiver(&fifo).unwrap();
+            barrier2.wait().await; // synchronize that both ends of the fifo file is open for writing and reading
+            let mut buf = [0; 64];
+            let read_count = reading_side.read(&mut buf).await.unwrap();
+            assert_eq!(read_count, 3);
+            assert_eq!(buf[0..3], vec![1, 2, 3]);
+        }
```
christoph-hamm: I would remove the block and instead use `drop(reading_side);`. This would be more obvious.

agent/doc/swdesign/README.md
```diff
@@ -3215,10 +3215,10 @@ Needs:
 
 Status: approved
 
-The Ankaios Agent shall ensure, that Control Interface input pipes are opened and messages are read.
+The Ankaios agent shall ensure, that Control Interface input pipes are opened and messages are read.
```
christoph-hamm, line 3218: I am not sure if we normally use the lowercase or uppercase here. Looking at the whole repository, we seem to use both version. I would use the uppercase version, as it is the name of the component.

christoph-hamm (changes requested):

agent/src/control_interface/control_interface_task.rs
```diff
@@ -100,9 +106,20 @@ impl ControlInterfaceTask {
         loop {
             select! {
                 // [impl->swdd~agent-ensures-control-interface-input-pipe-read~1]
-                from_server = self.input_pipe_receiver.recv() => {
+                from_server = self.from_server_receiver.recv() => {
                     if let Some(FromServer::Response(response)) = from_server {
-                        let _ = self.forward_from_server(response).await;
+                        if let Err(DeliveryError::NoReader(response)) = self.forward_from_server(response).await {
```
christoph-hamm: I do not like that the actual action, forwarding the message, is hidden on the right hand side of some nested if. It was hard for me to find, were the data is forwarded.

agent/doc/swdesign/README.md
krucod3: I was decided by team (while I was on vacations) to use lower case.

agent/src/control_interface/output_pipe.rs
krucod3: The information loss is intended here. If one does not need the `OutputPipeError` it is converted back to the internal type.

agent/src/control_interface/output_pipe.rs
krucod3: done.

agent/src/control_interface/output_pipe.rs
krucod3: Thx for the hint, we can refactor it with a for loop 👍

agent/src/control_interface/output_pipe.rs
krucod3: Copy-paste engineering 🤣 
I'll also update the old test that use blocks instead of drops 😉

agent/src/control_interface/control_interface_task.rs
krucod3: Good point 👍 Thx!

agent/src/control_interface/output_pipe.rs
krucod3: As discussed together, it does not work this way. Leaving as is.

agent/src/control_interface/output_pipe.rs
krucod3: As discussed together, it does not work this way. Leaving as is.

PR #514, opened 2025-07-07
LogsRequest server validation and LogsRequestAccepted response

GabyUnalaq:
Issues: #90

This PR contains 2 subtasks:
- Check the workload instance name on the server, upon receiving the LogsRequest.
- Send back a message upon accepting the request informing the user of the valid instance names.

# Definition of Done


api/proto/ank_base.proto
```diff
@@ -62,6 +63,13 @@ message LogsRequest {
 message LogsCancelRequest {
 }
 
+/**
+* A message indicating that the logs were successfully requested and for which workloads.
+*/
+message LogsRequestAccepted {
+    repeated WorkloadInstanceName workloadNames = 1; /// The names of the workloads for which the logs were successfully requested.
```
krucod3:
```suggestion
    repeated WorkloadInstanceName workloadNames = 1; /// The instance names of the workloads for which the logs request was successfully accepted.
```

server/src/ankaios_server.rs
```diff
@@ -339,10 +340,51 @@ impl AnkaiosServer {
...
+                            log::warn!(
+                                "No valid workload names provided in logs request: {:?}",
+                                logs_request.workload_names
+                            );
+                            self.to_agents
+                                .error(request_id, "No valid workload names provided".into())
+                                .await
+                                .unwrap_or_illegal_state();
+                            continue;
+                        }
```
krucod3: I would not mark this as an error. We provide a response already and this enables us to just return an empty list and let this be handled by the requesting side.

server/src/ankaios_server.rs
```diff
@@ -339,10 +340,51 @@ impl AnkaiosServer {
...
                         self.to_agents
-                            .logs_request(request_id.clone(), logs_request.into())
+                            .logs_request(request_id.clone(), valid_logs_request.clone().into())
+                            .await
+                            .unwrap_or_illegal_state();
+
+                        self.to_agents
+                            .logs_request_accepted(request_id.clone(), valid_logs_request.into())
                             .await
                             .unwrap_or_illegal_state();
```
krucod3: If we just provide the workload names here, we can spare the building of a new object on lines 376-379.

server/src/ankaios_server/server_state.rs
```diff
@@ -180,6 +181,21 @@ impl ServerState {
             .collect()
     }
 
+    pub fn get_workload_spec_by_instance_name(
+        &self,
+        instance_name: &WorkloadInstanceName,
+    ) -> Option<WorkloadSpec> {
```
krucod3: This is now returning clones of objects (that could potentially be big in case of file mounts without configs) that get discarded directly after the ownership transfer.

api/proto/ank_base.proto
```diff
@@ -40,8 +40,9 @@ message Response {
         Error error = 3;
         CompleteState completeState = 4;
         UpdateStateSuccess UpdateStateSuccess = 5;
-        LogEntriesResponse logEntriesResponse = 6; /// A message containing workload logs.
-        LogsStopResponse logsStopResponse = 7; /// A message containing the workload instance name indicating the stop of the log streaming.
+        LogsRequestAccepted logsRequestAccepted = 6; /// A message containing the workload names for which the logs were successfully requested.
```
krucod3, line 43: It's always bad to change the indices. In this we still work on an unreleased feature so we can afford it for the sake of ordering, but we will at least get in a conflict with the changes in #525.

krucod3 (commented):

api/proto/ank_base.proto
krucod3: Merged.

server/src/ankaios_server.rs
krucod3: fixed.

server/src/ankaios_server/server_state.rs
krucod3: fixed.

server/src/ankaios_server.rs
krucod3: fixed.

server/src/ankaios_server.rs
```diff
@@ -1354,18 +1327,25 @@ mod tests {
...
                         == &WorkloadInstanceName::new(AGENT_A, WORKLOAD_NAME_1, INSTANCE_ID)
                 },
             ))
             .once()
-            .return_const(None);
+            .return_const(false);
 
         server.server_state = mock_server_state;
 
+        server
```
inf17101, line 1370: Do we really need to insert a log campaign if we have an empty list of workloads after the retain logic?

inf17101 (commented):
Reviewed the changes from @krucod3. Added just one small question.

server/src/ankaios_server.rs
krucod3: Good point. The campaign would just be empty in this case.

inf17101: Then we can remove the `insert_log_campaign` call in utest `utest_server_forward_logs_request_invalid_workload_names`. But I noticed that in this utest the expectation counter is not checked, because I have expected that this utest would fail if the insert happens now only when the list after the retain is not empty. I checked out locally and I think we have this tokio task issue with the expectation counter of the mock we had already in the past. The expect call is useless. I have refactored the utest to consider the expectation counter properly:

Now, we can expect `never` called for invalid workload names.
```
    #[tokio::test]
    async fn utest_server_forward_logs_request_invalid_workload_names() {
        let _ = env_logger::builder().is_test(true).try_init();
        let (to_server, server_receiver) = create_to_server_channel(common::CHANNEL_CAPACITY);
        let (to_agents, mut comm_middle_ware_receiver) =
            create_from_server_channel(common::CHANNEL_CAPACITY);

        let mut server = AnkaiosServer::new(server_receiver, to_agents);
        let mut mock_server_state = MockServerState::new();

        mock_server_state
            .expect_desired_state_contains_instance_name()
            .with(mockall::predicate::eq(WorkloadInstanceName::new(
                AGENT_A,
                WORKLOAD_NAME_1,
                INSTANCE_ID,
            )))
            .once()
            .return_const(false);

        server.server_state = mock_server_state;

        server
            .log_campaign_store
            .expect_insert_log_campaign()
            .never();

        let logs_request = LogsRequest {
            workload_names: vec![WorkloadInstanceName::new(
                AGENT_A,
                WORKLOAD_NAME_1,
                INSTANCE_ID,
            )],
            follow: true,
            tail: 10,
            since: None,
            until: None,
        };

        // send logs request to server
        let logs_request_result = to_server
            .logs_request(REQUEST_ID.to_string(), logs_request)
            .await;
        assert!(logs_request_result.is_ok());

        assert!(to_server.stop().await.is_ok());

        let server_result = server.start(None).await;

        let from_server_command = comm_middle_ware_receiver.recv().await.unwrap();
        assert_eq!(
            from_server_command,
            FromServer::Response(ank_base::Response {
                request_id: REQUEST_ID.to_string(),
                response_content: Some(ank_base::response::ResponseContent::LogsRequestAccepted(
                    ank_base::LogsRequestAccepted {
                        workload_names: vec![],
                    }
                )),
            })
        );

        assert!(comm_middle_ware_receiver.try_recv().is_err());
        assert!(server_result.is_ok());
    }
```

inf17101: Utest is fine now.

inf17101 (approved):
LGTM 👍

PR #523, opened 2025-07-10
Reset agent log subscription store when server is gone

inf17101:
Issues: #90 

When the connection between server and agent is interrupted, the agent must delete all its internal subscriptions and stop the collecting logs of its workloads. This is why the internal `ServerGone` message is introduced with this PR to notify the agent about the connection interruption, so that it can delete all entries from its subscription store. The new message is not forwarded over network, just internally.

# Definition of Done


agent/src/agent_manager.rs
```diff
@@ -224,6 +224,16 @@ impl AgentManager {
                     .delete_subscription(&request_id);
                 Some(())
             }
+            FromServer::ServerGone => {
+                log::debug!("Agent '{}' received ServerGone.", self.agent_name);
```
krucod3:
```suggestion
                log::info!("Agent '{}' received ServerGone.", self.agent_name);
```

agent/src/subscription_store.rs
```diff
@@ -53,6 +53,10 @@ impl SubscriptionStore {
         self.store.remove(id);
     }
 
+    pub fn delete_all_subscriptions(&mut self) {
+        self.store.clear();
```
krucod3, line 58: We could add a debug trace here.

inf17101: Added in [e2c5954](https://github.com/eclipse-ankaios/ankaios/pull/523/commits/e2c5954187f4d1e262b86db214385db0dd7cf5ff)

PR #529, opened 2025-07-15
Handle LogsRequestAccepted messages in Ank CLI

inf17101:
Issues: #90 

The PR implements the handling of the `LogsRequestAccepted` message in the Ank CLI. However, the `LogsCancelRequestAccepted` message is not handled, because if a user terminates the CLI or closes the terminal window, the CLI would has to wait for the default timeout listening for the cancel accepted message which would lead to unnecessary blocking of the process without any significant benefit.

# Definition of Done

### DIFF
--- a/ank/src/cli_commands/server_connection.rs
+++ b/ank/src/cli_commands/server_connection.rs
@@ -19,7 +19,7 @@ use crate::filtered_complete_state::FilteredCompleteState;
 use crate::{output_and_error, output_debug};
 use std::{collections::BTreeSet, mem::take, time::Duration};
 
-use api::ank_base;
+use api::ank_base::{self, LogsRequestAccepted};
 use common::{
     commands::{CompleteStateRequest, LogsRequest, UpdateWorkloadState},
     communications_client::CommunicationsClient,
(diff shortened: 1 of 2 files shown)


ank/src/cli_commands/server_connection.rs
```diff
@@ -269,6 +276,55 @@ impl ServerConnection {
...
+                    )),
+            }))) => {
+                output_debug!(
+                    "LogsRequest accepted of request id '{}' for the following workload instance names: {:?}",
+                    request_id,
+                    logs_request_accepted_response.workload_names
+                );
+                Ok(logs_request_accepted_response)
+            }
+            Ok(Some(message)) => Err(ServerConnectionError::ExecutionError(format!(
```
krucod3: Maybe it would be better to just ignore the message here instead of failing. It could be that a workload state comes in between, for example, making the request fail without an obvious for the user reason.

The only question is if we want to store the message:
```rust
                        // [impl->swdd~cli-stores-unexpected-message~1]
                        self.missed_from_server_messages.push(message);
```

From what I saw, currently there would be no handling of the missed messages afterwards, so in theory we can spare it. On the other hand, storing the message is the right thing to do and could reduce confusion if need it later ...

inf17101: I ignore this message. But like discussed we avoid using the queue with missed messages at the moment. I put a comment with the reason into the code.

PR #540, opened 2025-07-22
Validate log rule workload names

christoph-hamm:
Issues: #90

# Definition of Done

### DIFF
--- a/common/src/objects/control_interface_access.rs
+++ b/common/src/objects/control_interface_access.rs
@@ -311,12 +311,6 @@ mod tests {
         const MAX_PREFIX: &str = "123456789012345678901234567890";
         const MAX_SUFFIX: &str = "123456789012345678901234567890123";
 
-        fn log_rule_with_workload(workload_name: &str) -> AccessRightsRule {
-            AccessRightsRule::LogRule(LogRule {
-                workload_names: vec![workload_name.to_string()],
-            })
-        }
-
         assert!(log_rule_with_workload("workload_1").verify_format().is_ok());
         assert!(log_rule_with_workload("*workload_1")
             .verify_format()
@@ -352,12 +346,6 @@ mod tests {
     fn utest_access_rights_log_rule_verify_fails() {
         const TOO_LONG_PREFIX: &str = "123456789012345678901234567890";
         const TOO_LONG_SUFFIX: &str = "1234567890123456789012345678901234";
-
-        fn log_rule_with_workload(workload_name: &str) -> AccessRightsRule {
-            AccessRightsRule::LogRule(LogRule {
-                workload_names: vec![workload_name.to_string()],
-            })
-        }
 
         assert!(log_rule_with_workload("").verify_format().is_err());
         assert!(
@@ -398,6 +386,12 @@ mod tests {
         assert!(log_rule_with_workload("also@wrong*")
             .verify_format()
             .is_err());
+    }
+
+    fn log_rule_with_workload(workload_name: &str) -> AccessRightsRule {
+        AccessRightsRule::LogRule(LogRule {
+            workload_names: vec![workload_name.to_string()],
+        })
     }
 
     // [utest->swdd~common-access-rules-filter-mask-convention~1]


common/doc/swdesign/README.md
```diff
@@ -445,6 +445,26 @@ Needs:
...
+
+Status: approved
+
+The Common library shall provide functionality for enforcing workload names in Control Interface access rules for logs:
+- to contain at most one "*" symbol
+- to be able to match a workload following the naming convention (e.g. not being to long, only containing valid characters)
+
+Rationale:
+This shall prevent users from providing rules which will never match any workload.
+Otherwise, invalid deny rules could lead to workloads having more rights than expected.
```
krucod3:
```suggestion
The Common library shall provide functionality for enforcing the validity of workload names in Control Interface access `LogRule`s:
- to contain at most one wildcard symbol "*"
- to be able to match a workload following the naming convention (e.g. not being to long, only containing valid characters)

Rationale:
This shall prevent users from providing rules which will never match any workload.
Otherwise, invalid deny rules could lead to workloads having more rights than expected.
```

common/src/objects/control_interface_access.rs
```diff
@@ -90,14 +97,42 @@ impl AccessRightsRule {
                 })?;
             }
             AccessRightsRule::LogRule(log_rule) => {
```
krucod3, line 97:
```suggestion
            // [impl->swdd~common-access-rules-logs-workload-names-convention~1]
            AccessRightsRule::LogRule(log_rule) => {
```

common/src/objects/workload_spec.rs
```diff
@@ -28,7 +28,7 @@ pub type WorkloadCollection = Vec<WorkloadSpec>;
 pub type DeletedWorkloadCollection = Vec<DeletedWorkload>;
 
 const MAX_CHARACTERS_WORKLOAD_NAME: usize = 63;
-pub const STR_RE_WORKLOAD: &str = r"^[a-zA-Z0-9_-]+$";
+pub const STR_RE_WORKLOAD: &str = r"^[a-zA-Z0-9_-]*$";
```
krucod3, line 32: With the current implementation we now allow empty names if only the regex is checked. This is not a problem at all, but we use the variable also in the CLI (apply_manifest.rs). 
We could make an edit of the trace there, but I find it in general bit dangerous to change the pattern. Maybe we can leave the pattern and just reorder the verification.

common/src/objects/control_interface_access.rs
```diff
@@ -90,14 +97,42 @@ impl AccessRightsRule {
...
+            let suffix = &workload_name[wildcard_pos + 1..];
+            if suffix.contains(WILDCARD_SYMBOL) {
+                Err(format!(
+                    "Expected at most one '{}' symbol.",
+                    WILDCARD_SYMBOL
+                ))
+            } else {
+                verify_workload_name_pattern(prefix)
+                    .and_then(|_| verify_workload_name_pattern(suffix))
+                    .and_then(|_| verify_workload_name_length(prefix.len() + suffix.len()))
```
krucod3, line 119: 🤣 We could add a +1 here as the wildcard should at the end match something, otherwise it is senseless.

common/src/objects/control_interface_access.rs
```diff
@@ -90,14 +97,42 @@ impl AccessRightsRule {
...
+            }
+        } else {
+            let length = workload_name.len();
+            verify_workload_name_format(workload_name)
+                .and_then(|_| verify_workload_name_length(length))
+                .and_then(|_| verify_workload_name_not_empty(length))
```
krucod3: Can't we directly call the `verify_workload_name_format()` here? The `map_err` can then be "LogRule error: {err}" (note, without ' ').

common/src/objects/control_interface_access.rs
```diff
@@ -272,42 +307,99 @@ mod tests {
...
-        assert!(AccessRightsRule::LogRule(LogRule {
-            workload_names: vec!["just.wrong".to_string()],
-        })
-        .verify_format()
-        .is_err());
+        fn log_rule_with_workload(workload_name: &str) -> AccessRightsRule {
+            AccessRightsRule::LogRule(LogRule {
+                workload_names: vec![workload_name.to_string()],
+            })
+        }
```
krucod3: Why not moving the function outside the test and using it in both tests?

common/src/objects/workload_spec.rs
krucod3: Discussed and decided to change the trace in the apply_manifets.rs to:
return Err("The provided path does not exist! This may be caused by invalid workload naming.".to_string());

common/src/objects/control_interface_access.rs
christoph-hamm: l would keep it as it is.

We allow wildcards to match the empty string (which is also something expected when we are using the "*" character). E.g. "ab*yz" would match the name "abyz". Hence "123456789012345678901234567890*1234567890123" could match "1234567890123456789012345678901234567890123". As it can match a valid workload name, it should be valid.

common/src/objects/control_interface_access.rs
christoph-hamm: I copied this part on purpose, as the resulting error is nicer. As this is some error users will see when using `ank apply`, I think the better error is more important, than 2 more lines of code.

common/src/objects/workload_spec.rs
christoph-hamm: I adapted the wording to always be "expected to have characters in {}" and introduced a new const `ALLOWED_SYMBOLS: &str = "[a-zA-Z0-9_-]"`.

common/src/objects/control_interface_access.rs
christoph-hamm: Done

krucod3 (approved):
LGTM 👍

PR #542, opened 2025-07-22
Add python sdk logging example

GabyUnalaq:
Issues: #90 

This PR adds a python_sdk example for the logging mechanism. In addition:
 - It improves the `run_example.sh` script, by being able to specify the manifest file used for the example.
 - It uses a special `setup_python_sdk.sh` script that sets the python sdk from a specific branch and extracts the protos from another (the ways to use it are in the script)
 - It adds another manifest file for creating another workload that logs a message every second.

# Definition of Done


examples/python_sdk/Dockerfile
```diff
@@ -0,0 +1,22 @@
+FROM docker.io/alpine:3.18.4 as base
+RUN apk update && apk add --update-cache \
+    # Development tools
+    python3 \
+    && rm -rf /var/cache/apk/* \
+    && python3 -m ensurepip && pip3 install protobuf==3.20.2
+
+RUN apk update && apk add --update-cache \
+    bash git curl \
+    && rm -rf /var/cache/apk/*
```
krucod3:
```suggestion
FROM docker.io/alpine:3.18.4 as base
RUN apk update && apk add --update-cache \
    # Development tools
    python3 \
    bash \
    git \
    curl \
    && rm -rf /var/cache/apk/* \
    && python3 -m ensurepip && pip3 install protobuf==3.20.2
```

examples/python_sdk/Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+
+COPY examples/python_sdk/setup_python_sdk.sh .
+
+# This will setup the python sdk from the "61_log_collection" branch
+# and will use the proto files from the "feature/90_provide_access_to_workload_logs" branch.
```
krucod3: Maybe the default should be main branch for both of them.
```suggestion
# This will setup the python sdk from the "main" branch of the ank-sdk-python
# and will use the proto files from the "main" branch of ankaios.
```

examples/python_sdk/Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+# Adjust the params as needed.
+RUN ./setup_python_sdk.sh --sdk-source github --sdk-branch 61_log_collection --proto-source branch --proto-branch feature/90_provide_access_to_workload_logs
```
krucod3:
```suggestion
RUN ./setup_python_sdk.sh --sdk-source github --sdk-branch main --proto-source branch --proto-branch main
```

examples/python_sdk/Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+
+COPY examples/python_sdk /ankaios
+ENV ANKAIOS_VERSION=0.6.0-pre
```
krucod3: We should update the version updating script to make changes here too.

examples/python_sdk/setup_python_sdk.sh
```diff
@@ -0,0 +1,156 @@
...
+    echo "Usage: $0 [--sdk-source <pypi|github|local>] [--sdk-version <version>] [--sdk-branch <branch>] [--proto-source <default|repo|branch|local>] [--proto-branch <branch>] [--proto-path <path>]"
+    echo ""
+    echo "If the installation is made from Pypi, the proto files will be fetched during the install with no possibility to change them."
+    echo ""
+    echo "Options:"
+    echo "  --help | -h        Show this help message and exit."
+    echo "  --sdk-source       Source of the SDK (pypi, github, local). Default: pypi."
+    echo "  --sdk-version      Version of the SDK (used with pypi)."
+    echo "  --sdk-branch       Branch to clone (used with github). Default: main."
+    echo "  --proto-source     Source of the proto files (branch, local). By default, it fetches them automatically."
```
krucod3: I suppose:
```suggestion
    echo "  --proto-source     Source of the proto files (branch, local). By default the files are fetched automatically with the SDK."
```

krucod3: I wonder if the local should not per default directly take the protos from the local repo. (default for `--proto-path`)

examples/python_sdk/setup_python_sdk.sh
```diff
@@ -0,0 +1,156 @@
+#!/bin/bash
```
krucod3: Exit if something goes wrong (e.g., git clone):
```suggestion
#!/bin/bash
set -e
```

We can probably also spare some error handling this way too (`curl -s ...`)

examples/python_sdk/src/main.py
```diff
@@ -0,0 +1,79 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+"""
+This example demonstrates how to request logs from workloads running on the system.
+It connects to the Ankaios control interface, retrieves the names of workloads,
+and requests logs for those workloads that match a specific name ("screamer" in this case).
+It then prints the logs received from the log campaign for a limited time (10 seconds).
+
+How to run:
+$ ./run_example.sh python_sdk --manifest-file logging_manifest.yaml
```
krucod3: I suppose we should document this in the README.md too.

In general it's a bit of a pitfall. I wounder if we should not use a dedicated manifest for each example. Now that we have the logs the user can without a problem request the logs for the example workload with `ank get logs <example name>`.

examples/python_sdk/Dockerfile
GabyUnalaq: After a discussion, we chose:
- the default should be the python sdk installed from pypi directly.
- If the user provides an environment variable called `SDK_BRANCH`, then the python sdk will be cloned from there and the proto files will be used from the current local version of ankaios.

I will add a `ReadMe.md` documenting this.

examples/python_sdk/Dockerfile
GabyUnalaq: Same situation as the finding above.

examples/python_sdk/setup_python_sdk.sh
GabyUnalaq: Because this script it is run inside the container, the proto files must first be copied onto the container. The location they are copied to is irrelevant but must be mentioned here.

examples/python_sdk/src/main.py
GabyUnalaq: I will add it in the designated `ReadMe.md` of the python_sdk example.

examples/python_sdk/Dockerfile
krucod3: The comment was targeting the update_version.sh, but it is already taking care of all docker files in the examples folder.

examples/README.md
```diff
@@ -1,20 +1,28 @@
...
-Every subfolder represents an example for a specific programming language. Feel free to try them out by running the steps described below.
+Every subfolder of the control interface examples represents one for a specific programming language. Feel free to try them out by running the steps described below.
 
 **Note:** The examples are simplified to focus on the usage of the control interface and designed for easy readability. They are not optimized for production usage.
 
+## Other examples
+
+Besides the control interface examples, there are custom ones added (and the list hopefully will grow):
+
+- python_sdk_logging: example uses the [Python SDK](https://pypi.org/project/ankaios-sdk/) to read the logs of another workload.
```
krucod3, line 24: We should add the sleepy here too.

examples/python_sdk_logging/README.md
```diff
@@ -0,0 +1,23 @@
...
+
+For testing the python SDK, you can specify a specific branch that the SDK should be installed from. This can be done by setting the `PYTHON_SDK_BRANCH` environment variable in the script:
+
+``` bash
+./run_example.sh python_sdk --env PYTHON_SDK_BRANCH=python_sdk_branch
+```
+
+This will also automatically use the proto files from the local ankaios repository.
+
+For more advanced configurations, you can modify the Dockerfile and changing directly the arguments for the `setup_python_sdk.sh` script.
```
krucod3:
```suggestion
For more advanced configurations, you can modify the Dockerfile or directly all the `setup_python_sdk.sh` script with the required arguments.
```

examples/python_sdk_logging/startupState.yaml
```diff
@@ -0,0 +1,26 @@
...
+      allowRules:
+        - type: StateRule
+          operation: Read
+          filterMask:
+            # workload has read access to the workload states of the cluster
+            - "workloadStates"
+          # Workload is allowed to read the logs of the workload screamer
+        - type: LogRule
+          workloadNames:
+            - screamer
```
krucod3:
```suggestion
          operation: ReadWrite
          filterMask:
            # workload has full access to the state of the cluster
            - "*"
          # Workload is allowed to read the logs of all workloads
        - type: LogRule
          workloadNames:
            - "*"
```

PR #556, opened 2025-07-31
Prepare main for development of 0.7.0-pre

krucod3:
Bumping the version to 0.7.0-pre

# Definition of Done

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -43,7 +43,7 @@ dependencies = [
 
 [[package]]
 name = "ank"
-version = "0.6.0"
+version = "0.7.0-pre"
 dependencies = [
  "api",
  "clap",
@@ -66,7 +66,7 @@ dependencies = [
 
 [[package]]
 name = "ank-agent"
-version = "0.6.0"
+version = "0.7.0-pre"
 dependencies = [
  "api",
  "async-trait",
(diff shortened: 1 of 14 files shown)


tools/ankaios-docker/agent/Dockerfile
```diff
@@ -1,7 +1,7 @@
 FROM ubuntu:24.04
 
 ARG TARGETARCH
-ARG VERSION=0.6.0
+ARG VERSION=0.7.0-pre
```
windsource: The `ankaios-docker` tool will not work that way as there is no asset for 0.7.0-pre to download. For that reason I suggest to revert that change.

Note that, the `update_version.sh` script has two modes. Only when called with `--release` - so for an officially released version - that version here will be updated.

tools/ankaios-docker/server/Dockerfile
```diff
@@ -1,7 +1,7 @@
 FROM ubuntu:24.04
 
 ARG TARGETARCH
-ARG VERSION=0.6.0
+ARG VERSION=0.7.0-pre
```
windsource: The `ankaios-docker` tool will not work that way as there is no asset for 0.7.0-pre to download. For that reason I suggest to revert that change.

Note that, the `update_version.sh` script has two modes. Only when called with `--release` - so for an officially released version - that version here will be updated.

windsource (changes requested):
Minor changes requested.

windsource (approved):
LGTM 👍

PR #566, opened 2025-08-13
Mock agent resource usage unit test

inf17101:
The utest `utest_agent_manager_sends_available_resources` sporadically fails because cpu usage is zero when running unit tests.
Instead of doing real system calls it shall be mocked.

This PR creates a mock version of the resource monitoring for unit tests.

# Definition of Done

### DIFF
--- a/agent/src/agent_manager.rs
+++ b/agent/src/agent_manager.rs
@@ -11,12 +11,10 @@
 // under the License.
 //
 // SPDX-License-Identifier: Apache-2.0
-use sysinfo::{CpuRefreshKind, MemoryRefreshKind, RefreshKind, System};
-
 use common::{
     commands::AgentLoadStatus,
     from_server_interface::{FromServer, FromServerReceiver},
-    objects::{CpuUsage, FreeMemory, WorkloadState},
+    objects::WorkloadState,
     std_extensions::{GracefulExitResult, IllegalStateResult},
     to_server_interface::{ToServerInterface, ToServerSender},
 };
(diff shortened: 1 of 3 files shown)


agent/src/resource_monitor.rs
```diff
@@ -0,0 +1,113 @@
...
+
+#[cfg(test)]
+use tests::MockSystem as System;
+
+pub struct ResourceMonitor {
+    refresh_kind: RefreshKind,
+    sys: System,
+}
+
+// [impl->swdd~agent-sends-node-resource-availability-to-server~1]
```
GabyUnalaq: The requirement states that the agent measures metrics and sends them. I am not sure if it makes sense here, and this applies to the utest link from this file as well.

GabyUnalaq (commented):
Just a small remark / question. The rest looks good to me.

HorjuRares (changes requested):
Taking into account the description of `swdd~agent-sends-node-resource-availability-to-server~1`:
```
At an interval of 2 seconds, the AgentManager measures the global CPU usage and the available free memory and sends them to the Ankaios server via an `AgentLoadStatus` message.
```
I think we either need to split the already existing requirements in 2 different ones (one which addresses the resource monitoring, and another one for the metrics sending) or update the existing one by mentioning that we are now delegating the sampling of metrics to a specialized struct used by the `AgentManager`.

agent/src/agent_manager.rs
```diff
@@ -636,6 +642,18 @@ mod tests {
...
+            .expect()
+            .once()
+            .return_once(|| {
+                let mut mock_resource_monitor = MockResourceMonitor::default();
+                mock_resource_monitor
+                    .expect_sample_resource_usage()
+                    .returning(|| (CpuUsage::new(50.0), FreeMemory { free_memory: 1024 }));
+                mock_resource_monitor
+            });
+
```
HorjuRares, line 656: I like this part really much as it takes advantage of benefits of mocking.
Also, besides the requirement issue, everythin looks fine for me 👍

GabyUnalaq:
> Taking into account the description of `swdd~agent-sends-node-resource-availability-to-server~1`:
> 
> ```
> At an interval of 2 seconds, the AgentManager measures the global CPU usage and the available free memory and sends them to the Ankaios server via an `AgentLoadStatus` message.
> ```
> 
> I think we either need to split the already existing requirements in 2 different ones (one which addresses the resource monitoring, and another one for the metrics sending) or update the existing one by mentioning that we are now delegating the sampling of metrics to a specialized struct used by the `AgentManager`.

For this PR to not extend further than it's intended fix, I believe we can just remove the requirement links in `resource_monitor.rs`, so that they are linked the same as before. If the requirement should be changed / improved / split, we can do this in another PR.

agent/src/resource_monitor.rs
inf17101: I think I will remove the linkage here.

inf17101: Removed in [0f73457](https://github.com/eclipse-ankaios/ankaios/pull/566/commits/0f73457c1489fd9d155724a276f43a31ebc51342)

GabyUnalaq (approved):
LGTM 👍

HorjuRares (approved):
Tanking into account possible future modifications and expanding of the resource measurement, the removal of linkage seems appropriate.

LGTM 👍

Issue #450, opened 2025-02-06
Events support


krucod3:
## Description

This issue is about adding "events" support to Ankaios. The general idea is that workloads can register for certain events and get notifications when the event occurs. Events can be of different types:
* Desired state changes
* Workload execution state changes
* Dynamic workload produced events

Note: If we add the dynamic events as part of the complete state, all types can be handled as complete state update events where the interested party registers with a filter on the complete state, e.g., `desiredState.workloads.infotainment_*`.

Events must be available through both the Control Interface available to workload as well as the Command Interface available to the the ank CLI and third-party applications.

Events can also be used as dependency triggers. An app can only be started if an event occurs.

## Goals

* Enable workload execution states pushing (without polling). Use-cases:
    * reporting problems to base (e.g., OEM cloud)
* Extend the feature set for the dependency handling. Use-case:
    * workload depend on an event <workload name>-ready that is sent by some sort of setup app (e.g., network manager). 
* Better support for Ankaios plugins that extend the functionality without changing the Ankaios codebase. Use-cases:
    * scheduling or network management plugins that get informed that a new unscheduled workload is available by registering for changes to `desiredState.workload`
* Better integration of third party tools running outside of/beside the cluster. Use-case:
    * control app that connects the edge device to fleet management in the cloud
* Enable inter-workload communication by allowing workloads to emit dynamic events that are read by other workloads. Use-case:
    * Complex app deployments where no heavy communication between the parts of the app is required and a small number of events is a better option than a dedicated network setup.

**Note:**
Especially network management use-cases require some additional changes in Ankaios, as even if the network is prepared beforehand, the runtime must be specifically instructed to use the prepared setup. This means that network related configs must be written somewhere extra s.t. they can be read by the runtime connectors. As this would be a build in feature, it would be better to use a dedicated optional networking attribute per workload which is transparent to Ankaios, but can be read by a runtime connector. The alternative of using the workload tags would somehow be an ugly patch.

### Summary

**To be filled when the final solution is sketched.**

### Tasks

- [ ] Task 1
- [ ] Task 2
- [ ] ...

GabyUnalaq:
After a discussion with @krucod3, we decided to leave the server disconnects case. Without a proper message informing the workload that the server is gone, we would need a specific message in the API for this, which would be overwritten by the server gone one. The current solution is to leave it as it is. Effect: The workload will expect events with a specific ID that will never come, because after a restart, the server will be unaware of the campaigns that were started before.

[...]

Issue #521, opened 2025-07-10
Agent disconnects must be forwarded to the Control Interface


krucod3:
## Description

If a workload communicates with Ankaios on the Control Interface and there is a short network hick-up between the server and the agent, messages sent to the workload could get lost.

This can include `UpdateStateSuccess`, but also log campaign canceling. In the case of the log campaign canceling the workload could wait forever for more messages from the observed workload.

The same problem could occur when events are added to Ankaios as planned with #450.

To handle the problem two messages need to be sent:
* server disconnected when the networking problem starts
* some reconnected message (could also be an additional hello which will be added with #453)

### Summary

**To be filled when the final solution is sketched.**

### Tasks

- [ ] Task 1
- [ ] Task 2
- [ ] ...

krucod3:
Not planned for now (see #541).

PR #573, opened 2025-08-29
Make changes to gRPC API for supporting events

christoph-hamm:
Issues: #450

# Definition of Done

### DIFF
--- a/ank/src/cli_commands/server_connection.rs
+++ b/ank/src/cli_commands/server_connection.rs
@@ -108,7 +108,7 @@ impl ServerConnection {
                 request_id.to_owned(),
                 CompleteStateRequest {
                     field_mask: object_field_mask.to_vec(),
-                    subscribe: false,
+                    subscribe_for_events: false,
                 },
             )
             .await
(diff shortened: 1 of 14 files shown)


api/proto/ank_base.proto
```diff
@@ -133,6 +134,18 @@ message UpdateStateSuccess {
     repeated string deletedWorkloads = 2; /// Workload instance names of workloads which will be stopped
 }
 
+
+/*** A message containing the complete state of the Ankaios system along with information about what has changed.
+* This is a response to the [CompleteStateRequest](#completestaterequest) message.
+* If this is the first response to a CompleteStateRequest,l the addedFields, updatedFields, and removedFields will all be empty.
```
inf17101:
```suggestion
* If this is the first response to a CompleteStateRequest, the addedFields, updatedFields, and removedFields will all be empty.
```

windsource:
We should not merge this to main but to a feature branch such that it only appears on main when also the implementation is existing.

api/proto/ank_base.proto
```diff
@@ -38,7 +38,7 @@ message Response {
     string requestId = 1;
     oneof ResponseContent {
         Error error = 3;
-        CompleteState completeState = 4;
+        CompleteStateResponse completeState = 4;
```
krucod3: We should call this `completeStateResponse` too as it is not a complete state only.

api/proto/ank_base.proto
```diff
@@ -133,6 +134,18 @@ message UpdateStateSuccess {
...
+
+/*** A message containing the complete state of the Ankaios system along with information about what has changed.
+* This is a response to the [CompleteStateRequest](#completestaterequest) message.
+* If this is the first response to a CompleteStateRequest,l the addedFields, updatedFields, and removedFields will all be empty.
+*/
+message CompleteStateResponse {
+    CompleteState completeState = 1; /// The complete state of the Ankaios system.
+    repeated string addedFields = 2; /// The fields that were added in the state compared to the previous state.
+    repeated string updatedFields = 3; /// The fields that were updated in the state compared to the previous state.
+    repeated string removedFields = 4; /// The fields that were removed in the state compared to the previous state.
```
krucod3: As discussed, maybe we can extract these to a dedicated message (as we could reuse it later in the `UpdateStateSuccess`).
Maybe something like:
* `AffectedFields`
* `TouchedFields`
* `InfluencedFields`

api/proto/ank_base.proto
christoph-hamm: Fixed

api/proto/ank_base.proto
```diff
@@ -112,6 +112,7 @@ enum AddCondition {
 */
 message CompleteStateRequest {
     repeated string fieldMask = 1; /// A list of symbolic field paths within the State message structure e.g. 'desiredState.workloads.nginx'.
+    bool subscribe = 2; /// If true, the server will send updates to the client whenever the state changes. The updates will contain only the fields specified in the fieldMask. If false, the server will send only one response.
```
krucod3: Maybe we can rename this to get a common notion:
```suggestion
    bool subscribeForEvents = 2; /// If true, the server will send updates to the client whenever the state changes. The updates will contain only the fields specified in the fieldMask. If false, the server will send only one response.
```

api/proto/ank_base.proto
```diff
@@ -38,7 +38,7 @@ message Response {
     string requestId = 1;
     oneof ResponseContent {
         Error error = 3;
-        CompleteState completeState = 4;
+        CompleteStateResponse completeStateResponse = 4;
```
krucod3, line 42: How are we going to cancel an events campaign?

We should add a new request - `EventsCancelRequest`
and a new response - `EventsCancelAccepted`.

We should also handle the use-case that the server disappears. Either by adding a special unsolicited response (way too much work) or adding the use-case to #521.

ank/src/cli_commands/server_connection.rs
```diff
@@ -734,34 +735,38 @@ mod tests {
...
+                                ]
+                                .into(),
+                            }),
+                            files: Some(generate_test_proto_workload_files()),
+                        },
+                    )])),
+                    ..Default::default()
+                },
+            )),
         );
```
krucod3: How about adding a `generate_test...` function for this? (Or use one if it existis)
Next time we change something we will at least spare ourselvs the impl and review effort a bit.

api/proto/ank_base.proto
```diff
@@ -133,6 +134,25 @@ message UpdateStateSuccess {
     repeated string deletedWorkloads = 2; /// Workload instance names of workloads which will be stopped
 }
 
+
+/*** A message containing the complete state of the Ankaios system along with information about what has changed.
```
GabyUnalaq:
```suggestion
/** A message containing the complete state of the Ankaios system along with information about what has changed.
```

ank/src/cli_commands/server_connection.rs
GabyUnalaq: Added a `generate_test_complete_state_response` that generates the needed `ResponseContent::CompleteStateResponse`.

api/proto/ank_base.proto
GabyUnalaq: Added `EventCancelRequest` and `EventCancelAccepted` as discussed.

api/proto/ank_base.proto
GabyUnalaq: Done.

common/src/commands.rs
```diff
@@ -94,13 +95,18 @@ impl From<RequestContent> for ank_base::request::RequestContent {
             RequestContent::UpdateStateRequest(content) => {
                 ank_base::request::RequestContent::UpdateStateRequest(Box::new((*content).into()))
             }
-            // TODO: tests are missing for the next two cases
+            // TODO: tests are missing for the next three cases
```
krucod3, line 98: We could have also just added the tests 🤣

PR #581, opened 2025-09-10
First implementation of wildcard support

christoph-hamm:
Issues: #450

# Definition of Done


common/src/state_manipulation/object.rs
```diff
@@ -278,6 +278,66 @@ impl Object {
...
+                    current_result_valid = false;
+                    match current_value {
+                        Value::Mapping(map) => {
+                            for (key, value) in map {
+                                let key = match key {
+                                    Value::String(s) => s.clone(),
+                                    Value::Number(n) if n.is_i64() || n.is_u64() => n.to_string(),
+                                    _ => continue,
+                                };
+                                // let Value::String(key) = key else { continue };
```
inf17101: As you cover the handling between number and string as key we can remove this comment out code.

common/src/state_manipulation/object.rs
```diff
@@ -278,6 +278,66 @@ impl Object {
...
+                            }
+                        }
+                        _ => {}
+                    }
+
+                    break;
+                } else {
+                    current_prefix.push(path_element.clone());
+                    current_value = if let Some(x) = current_value.get(path_element) {
+                        x
```
inf17101:
```suggestion
                    current_value = if let Some(next_element) = current_value.get(path_element) {
                        next_element
```

common/src/state_manipulation/object.rs
```diff
@@ -278,6 +278,66 @@ impl Object {
...
+                                to_do.push_front((new_prefix, value, remaining_path));
+                            }
+                        }
+                        Value::Sequence(seq) => {
+                            for (index, value) in seq.iter().enumerate() {
+                                let mut new_prefix = current_prefix.clone();
+                                new_prefix.push(index.to_string());
+                                to_do.push_front((new_prefix, value, remaining_path));
+                            }
+                        }
```
inf17101: The wildcard support for sequences is nice, but like discussed in general we do not support currently filter for a specific sequence entry like this `ank get state desiredState.workloads.nginx.tags.0`. In this case just apiVersion is output from the state. Maybe we can fix it so that this sequence filtering you have implemented is also supported.

common/src/state_manipulation/object.rs
```diff
@@ -278,6 +278,66 @@ impl Object {
         Some(current_obj)
     }
 
+    pub fn expand_wildcards(&self, path: &[Path]) -> Vec<Path> {
```
inf17101, line 316: For tests I use the default server manifest `server/resources/startConfig.yaml`.
In general, the algorithm works pretty fine, however, when the user filters too much levels with wildcards like 
`ank get state desiredState.workloads.*.*.*` then I get a strange filtered output like this:
```yaml
desiredState:
  apiVersion: v0.1
  workloads:
    hello-pod:
      files: []
    hello1:
      dependencies:
        filesystem_init: ADD_COND_SUCCEEDED
      controlInterfaceAccess:
        allowRules:
        - type: StateRule
          operation: Read
          filterMask:
          - desiredState
        denyRules: []
      files: []
    hello2:
      files: []
    hello_containerd:
      files: []
    nginx:
      files: []
```

In this case the result vector contains mask parts with duplicated keys like `["desiredState", "workloads", "nginx", "files", "files"]`. Maybe this is an indicator for the problem.

When adding another invalid level with an wildcard `ank get state desiredState.workloads.*.*.*.*`, then I get an empty state output `{}`. Normally, just the apiVersion shall be output in this case.

inf17101 (commented):
Finished review.
Swdds are missing for now. We can add them also in an extra PR when the feature is ready. But personally, I would prefer to add them in the same PR.

Maybe one system test would be also good.

common/src/state_manipulation/object.rs
```diff
@@ -278,6 +278,66 @@ impl Object {
...
+        let value = &self.data;
+        let mut result = Vec::new();
+        let mut to_do = path
```
inf17101, line 319: Maybe a doc string for this method with the algorithm (DFS like impl) mentioned would be good.

common/src/state_manipulation/object.rs
```diff
@@ -672,6 +734,118 @@ mod tests {
...
+        };
+
+        let expaned = data.expand_wildcards(&["A.a.age".into(), "B.b.name".into()]);
+
+        let x = expaned
+            .iter()
+            .map(ToString::to_string)
+            .collect::<HashSet<_>>();
+
+        assert_eq!(x, HashSet::from(["A.a.age".into(), "B.b.name".into(),]));
```
inf17101: Can extract this to a variable to not repeat the input and the expected output (at least the inner array).

common/src/state_manipulation/object.rs
```diff
@@ -278,6 +278,66 @@ impl Object {
...
+                        break;
+                    }
+                }
+            }
+            if current_result_valid {
+                result.push(current_prefix);
+            }
+        }
+
+        result.into_iter().map(|p| p.into()).collect()
```
inf17101: Can we directly fill a `Vec<Path>` avoiding the conversion code at the end?

common/src/state_manipulation/object.rs
christoph-hamm: I tried to add general support for filtering in sequences, but this is to much effort. Especially as we plan to change tags to a map, which would only leave file mounts and access rights using sequence. For those two I do not see a use case in subscription or filtering.

I will remove the code handling sequences here.

common/src/state_manipulation/object.rs
christoph-hamm: Most of the observation have been artifacts of sequence. As I will remove the sequence support in the wildcard expansion, as stated above, this problems will disappear. For the `... "files", "files"]`, this is a result of out data structures and I have to take a closer look.

common/src/state_manipulation/object.rs
christoph-hamm: fixed

common/src/state_manipulation/object.rs
christoph-hamm: I have removed the code for sequences.

common/src/state_manipulation/object.rs
christoph-hamm: With the current version for `ank get state 'desiredState.workloads.*.*.*.*'` you get:

```
desiredState:
  apiVersion: v0.1
  workloads:
    hello1:
      dependencies:
        filesystem_init: ADD_COND_SUCCEEDED
      controlInterfaceAccess:
        allowRules:
        - type: StateRule
          operation: Read
          filterMask:
          - desiredState
        denyRules: []
```

I think this is the expected output.

common/src/state_manipulation/object.rs
christoph-hamm: I have added a doc string.

common/src/state_manipulation/object.rs
christoph-hamm: Fixed.

common/src/state_manipulation/object.rs
christoph-hamm: Yes, that is better. Good catch. I only had to change two lines.

christoph-hamm:
@inf17101, I have fixed the finding. Could you review my changes?

common/src/state_manipulation/object.rs
inf17101: Good that you removed it. But just for correctness here. The output of `ank get state 'desiredState.workloads.*.*.*.*'` for me is the expected one: 

Filter level too deep => only apiVersion is output
```yaml
desiredState:
  apiVersion: v0.1
```

The filter mask with 3 wildcards leads to the following, which seems to be unexpected output to me:

This is a little bit strange. First wildcard means all workload names, second wildcard means all fields of those workloads, but the third one produces just an output of `hello1`.
```yaml
desiredState:
  apiVersion: v0.1
  workloads:
    hello1:
      dependencies:
        filesystem_init: ADD_COND_SUCCEEDED
      controlInterfaceAccess:
        allowRules:
        - type: StateRule
          operation: Read
          filterMask:
          - desiredState
        denyRules: []
```

common/src/state_manipulation/object.rs
```diff
@@ -278,6 +278,59 @@ impl Object {
         Some(current_obj)
     }
 
+    /// Expands wildcard paths into all concrete paths present in the object.
+    ///
+    /// For each input path, this method performs a depth-first search, replacing any wildcard segment (e.g., "*") with all available keys at that level.
+    /// Returns a vector containing every concrete path that matches the provided wildcard patterns.
```
inf17101:
```suggestion
    /// Expands wildcard paths into all concrete paths present in the object.
    ///
    /// For each input path, this method performs a depth-first search, replacing any wildcard segment (e.g., "*") with all available keys at that level.
    /// ## Arguments
    ///
    /// - `path`: A slice of [`Path`] that may contain wildcards
    ///
    /// ## Returns
    ///
    /// - a [Vec<`Path`>] containing every concrete path that matches the provided wildcard patterns.
    ///
```

inf17101 (commented):
I have only one suggestion to the doc string fitting better to the rust conventions.

common/src/state_manipulation/object.rs
inf17101: I have seen it. I have made an suggestion above to add sections like in Rust doc string guidelines.

common/src/state_manipulation/object.rs
christoph-hamm: The third level are are `filesystem_init`, `allowRules` and `denyRules`. As only the workload `hello1` has `dependencies` and `controlInterfaceAccess`, only this workload is returned.

common/doc/swdesign/README.md
```diff
@@ -535,6 +535,96 @@ Needs:
...
+Needs:
+- impl
+- utest
+
+#### State manipulation provides functionality to expand wildcards
+`swdd~common-state-manipulation-expand-wildcards~1`
+
+Status: approved
+
+The Common library provides a methods for expanding paths containing wildcards ('*') as segments, using an exisiting object.
```
inf17101:
```suggestion
The Common library shall provide a method for expanding paths containing wildcards ('*') as segments, using an existing object.

Comment:
The result is a list of all paths valid for the given object, resulting from replacing wildcard segments with any possible segment.
```

common/doc/swdesign/README.md
```diff
@@ -535,6 +535,96 @@ Needs:
...
+Needs:
+- impl
+- utest
+
+#### State manipulation allows to set values
+`swdd~common-state-manipulation-set~1`
+
+Status: approved
+
+The Common library provides a method to set the value of an object at a certain path.
```
inf17101:
```suggestion
The Common library shall provide a method to set the value of an object at a certain path.
```

common/doc/swdesign/README.md
```diff
@@ -535,6 +535,96 @@ Needs:
...
+- impl
+- utest
+
+#### State manipulation set operation adds missing objects
+`swdd~common-state-manipulation-set-add-missing-objects~1`
+
+Status: approved
+
+When setting the value of an object at a certain path, the Common library adds missing intermediate objects.
+This intermediate objects are empty, except the newly added value.
```
inf17101:
```suggestion
When setting the value of an object at a certain path, the Common library shall add missing intermediate objects as empty objects followed by the newly added value.
```

common/doc/swdesign/README.md
```diff
@@ -535,6 +535,96 @@ Needs:
...
+Needs:
+- impl
+- utest
+
+#### State manipulation allows to remove values
+`swdd~common-state-manipulation-remove~1`
+
+Status: approved
+
+The Common library provides a method to remove the value of an object at a certain path.
```
inf17101:
```suggestion
The Common library shall provide a method to remove the value of an object at a certain path.
```

common/doc/swdesign/README.md
```diff
@@ -535,6 +535,96 @@ Needs:
...
+Needs:
+- impl
+- utest
+
+#### State manipulation allows to get values
+`swdd~common-state-manipulation-get~1`
+
+Status: approved
+
+The Common library provides a method to get the value of an object at a certain path.
```
inf17101:
```suggestion
The Common library shall provide a method to get the value of an object at a certain path.
```

common/doc/swdesign/README.md
```diff
@@ -535,6 +535,96 @@ Needs:
...
+### State manipulation
+
+Provides methods for accessing or updating parts of objects, as used by field masks.
+
+#### State manipulation uses period separated paths
+`swdd~common-state-manipulation-path~1`
+
+Status: approved
+
+The state manipulation methods of the Common library use paths separated by the '.' symbols.
```
inf17101:
```suggestion
The state manipulation methods of the Common library shall use paths separated by the '.' symbols.
```

inf17101 (commented):
Finished with reviews. When the merge conflict is resolved, we can merge the PR.

PR #585, opened 2025-09-10
Update to new tonic and prost versions

krucod3:
Issues: #

Use new tonic and prost versions

# Definition of Done

### DIFF
--- a/grpc/src/grpc_agent_connection.rs
+++ b/grpc/src/grpc_agent_connection.rs
@@ -13,7 +13,6 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use std::pin::Pin;
-use std::sync::Arc;
 
 use common::check_version_compatibility;
 use common::std_extensions::GracefulExitResult;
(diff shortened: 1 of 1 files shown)


grpc/src/grpc_agent_connection.rs
```diff
@@ -49,7 +50,7 @@ impl GRPCAgentConnection {
     }
 }
 
-fn has_multiple_peer_certs(peer_certs: &[Certificate]) -> bool {
+fn has_multiple_peer_certs(peer_certs: &Arc<Vec<CertificateDer>>) -> bool {
```
christoph-hamm: Are you sure you need this `Arc` here? For me it also builds with `peer_certs: &[CertificateDer]`. You could then also remove the `use std::sync::Arc;` above.

krucod3: You are right, a double implicit deref does the trick here.
I'll change it back.

PR #606, opened 2025-10-17
Support events in control interface tester

christoph-hamm:
Issues: #450

Add support for event subscription in control interface tester.
This changes allow to write system tests for event subscription.

# Definition of Done

The PR shall be merged only if all items mentioned in [CONTRIBUTING.md](https://github.com/eclipse-ankaios/ankaios/blob/main/CONTRIBUTING.md#how-to-contribute) have been followed. In case an item is not applicable as described, please provide a short explanation in the description.

- [ ] All steps in [CONTRIBUTING.md](https://github.com/eclipse-ankaios/ankaios/blob/main/CONTRIBUTING.md#how-to-contribute) have been handled

PR #611, opened 2025-10-23
Add event handler

inf17101:
Issues: #450 

This event handler implementation is based on the field comparison integration for the desiredState in PR #605.
The PR #605 is currently under review and must be merged first into the feature branch. Then this branch needs to be updated so that all changes are on top of it.

Changes:

- Add event handler implementation
- Move common `RequestId` parsing for agents and ank-cli from `log_campaign_store.rs` into an own sub module, because event handler also uses request id types as key in its internal event storage
- Add utests for event handler
- Add swdds for event handler and for Ankaios Server (including swdd for `EventsCancelAccepted` message)
- Moves the state comparison into the `StateComparator` struct and move the module to the server part because of mocking library availability
- Integrate event handler into the `ankaios_server.rs` in multiple places where events need to be checked and send out:
    1. DesiredState + Pending(Initial) workload states on startup with manifest
    2. UpdateStateRequest: DesiredState changes
    3. UpdateWorkloadState: New workload states
    4. AgentHello: new agent added
    5. AgentGone: agent removed + AgentDisconnected workload states as events
    6. AgentLoadStatus: New agent load status (Cpu, Memory) to be send out as events (while for agent and workload states an added and updated field is determined, for simplicity every new agent load state is sent out as an updated field (otherwise you need to compare the floating point numbers and return an enum communicating to the server what field has been added or updated and this seems to be impractical). Since the default floating point values are added to the resource availability in case of an AgentHello, the AgentLoadStatus messages are anyway preceeding updates only.

The event handler is only invoked if there are subscribers to event.

Stests are added in a seaparte PR when the PR #606 is merged to support stests for the control interface tester workload.
Also handling special cases for removing event subscriptions from the event handler are added in a separate PR.

An end-to-end test were made with different scenarios using the prepared PR https://github.com/eclipse-ankaios/ank-sdk-rust/pull/36 in ank-sdk-rust using the `state_event.rs` example app.
# Definition of Done

### DIFF
--- a/server/src/ankaios_server/request_id.rs
+++ b/server/src/ankaios_server/request_id.rs
@@ -22,12 +22,12 @@ pub type CliConnectionName = String;
 #[derive(Debug, Clone, PartialEq, Eq, Hash)]
 pub struct CliRequestId {
     pub cli_name: CliConnectionName,
-    pub request_uuid: String,
+    pub request_id: String,
 }
 
 impl Display for CliRequestId {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        write!(f, "{}@{}", self.cli_name, self.request_uuid)
+        write!(f, "{}@{}", self.cli_name, self.request_id)
     }
 }
 
@@ -35,7 +35,7 @@ impl Display for CliRequestId {
 pub struct AgentRequestId {
     pub agent_name: AgentName,
     pub workload_name: WorkloadName,
-    pub request_uuid: String,
+    pub request_id: String,
 }
 
 pub fn to_string_ids<I>(request_ids: HashSet<I>) -> HashSet<String>
@@ -57,7 +57,7 @@ impl Display for AgentRequestId {
         write!(
             f,
             "{}@{}@{}",
-            self.agent_name, self.workload_name, self.request_uuid
+            self.agent_name, self.workload_name, self.request_id
         )
     }
 }
@@ -103,7 +103,7 @@ where
                 .collect();
             RequestId::CliRequestId(CliRequestId {
                 cli_name: parts[CLI_REQUEST_NAME_INDEX].to_string(),
-                request_uuid: parts[CLI_REQUEST_ID_INDEX].to_string(),
+                request_id: parts[CLI_REQUEST_ID_INDEX].to_string(),
             })
         } else {
             let parts: Vec<&str> = request_id
@@ -113,7 +113,7 @@ where
             RequestId::AgentRequestId(AgentRequestId {
                 agent_name: parts[AGENT_REQUEST_NAME_INDEX].to_string(),
                 workload_name: parts[AGENT_REQUEST_WORKLOAD_NAME_INDEX].to_string(),
-                request_uuid: parts[AGENT_REQUEST_ID_INDEX].to_string(),
+                request_id: parts[AGENT_REQUEST_ID_INDEX].to_string(),
             })
         }
     }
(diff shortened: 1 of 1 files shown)


common/src/state_manipulation/object.rs
```diff
@@ -135,6 +132,15 @@ impl TryInto<proto::CompleteState> for Object {
     }
 }
 
+impl From<Object> for serde_yaml::Mapping {
```
christoph-hamm: This is not how the `From` trait is indented. The use of `from` should not lead to the loss of data.

I see why this implementation is appealing, as it saves work when using it.
I think implementing `TryFrom` and using `try_from().unwrap_or_default()` would be more idiomatic and the code using it would be clearer.

inf17101:
@christoph-hamm: Like discussed you can already start to review. I have resolved the branch conflicts and now I will add system tests.

server/src/ankaios_server.rs
```diff
@@ -203,12 +234,38 @@ impl AnkaiosServer {
...
+                        let altered_fields = vec![FieldDifference::removed_agent(&agent_name)];
+                        let altered_fields = agent_workload_states.iter().fold(
+                            altered_fields,
+                            |mut altered_fields, ws| {
+                                altered_fields.push(FieldDifference::updated_workload_state(
+                                    &ws.instance_name,
+                                ));
+                                altered_fields
+                            },
+                        );
```
christoph-hamm: I would replace the fold with an extend. Something like this (Caution, I have not tested this code):

```suggestion
                        let mut altered_fields = vec![FieldDifference::removed_agent(&agent_name)];
                        altered_fields.extend(
                            agent_workload_states.iter().map(|ws| FieldDifference::updated_workload_state(&ws.instance_name))
                        );
```

christoph-hamm (changes requested):

server/src/ankaios_server.rs
inf17101: This is outdated now since the code has changed again.

server/src/ankaios_server/request_id.rs
```diff
@@ -0,0 +1,160 @@
...
+use std::fmt::{self, Display};
+
+pub type AgentName = String;
+pub type WorkloadName = String;
+pub type CliConnectionName = String;
+
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub struct CliRequestId {
+    pub cli_name: CliConnectionName,
+    pub request_uuid: String,
```
christoph-hamm: Why are you calling it UUID? It is just some ID.

server/src/ankaios_server/state_comparator.rs
```diff
@@ -0,0 +1,885 @@
...
+    pub fn is_empty(&self) -> bool {
+        self.added_tree.is_empty() && self.removed_tree.is_empty() && self.updated_tree.is_empty()
+    }
+
+    fn insert_path(tree: &mut Object, at_path: Path, new_value: Value) {
+        if at_path.parts().is_empty() {
+            return;
+        }
+
+        tree.set(&at_path, new_value).unwrap_or_illegal_state();
```
christoph-hamm: I would suggest you remove `insert_added_path` and `insert_removed_path`. In the production code both function are only used for altering `first_difference_tree`.
* The second parameters is not used in productions
* The altering of `full_difference_tree` is done with `insert_path` anyhow. 
Just using the `insert_path` would be more consistent.

server/src/ankaios_server/state_comparator.rs
```diff
@@ -0,0 +1,885 @@
...
+        };
+
+        let mut state_difference_tree = StateDifferenceTree::new();
+        let mut stack_tasks = Vec::new();
+
+        stack_tasks.push(StackTask::VisitPair(&self.old_state, &self.new_state));
+        let mut current_field_mask = Vec::new();
+        while let Some(task) = stack_tasks.pop() {
+            match task {
+                StackTask::VisitPair(current_node, other_node) => {
```
christoph-hamm: I would stick to the terms "old" and "new", as their meaning is obvious. Especially the term `other` is ambiguous.

common/src/state_manipulation/object.rs
inf17101: Fixed in [e008d08](https://github.com/eclipse-ankaios/ankaios/pull/611/commits/e008d0877769476de8fe51a6042524f1292dd250) and [90b0a5a](https://github.com/eclipse-ankaios/ankaios/pull/611/commits/90b0a5a1417244535cf8d61fd6a4aedfb98f88fd).

server/doc/swdesign/README.md
```diff
@@ -1290,6 +1273,224 @@ Needs:
...
+`swdd~server-calculates-state-differences~1`
+
+Status: approved
+
+The StateComparator shall provide a method for calculating the state differences between a current state and a new state represented as associative data structures with returning tree structures for added, updated and removed field paths.
+
+Comment:
+Each tree contains the paths of the fields that have been changed, with a null value as the leaf.
+A custom Depth-Search-First (DFS) implementation comparing all fields of the current and new state is used.
+A sequence is treated as a leaf and in mappings only `string`s and numbers are supported as keys.
```
christoph-hamm: Split this into two points

server/doc/swdesign/README.md
```diff
@@ -1290,6 +1273,224 @@ Needs:
...
+For added workloads, only the field path to the newly added workload name is required in the altered event fields for subscribers, rather than the full path to each added sub-field of the new workload. The same rationale applies to workloads that have been removed.
+
+Tags:
+- StateComparator
+
+Needs:
+- impl
+- utest
+
+#### Server sends state differences as events
```
christoph-hamm: What is about updated field paths? Is the tree until the first level of change or the full tree generated?

server/src/ankaios_server/request_id.rs
inf17101: Like said, I just moved it to a new location and the names were existing already. I have renamed it.

server/src/ankaios_server/state_comparator.rs
inf17101: Yes the methods are there because of previous usage. They are now removed and migrated to `insert_path`.

server/src/ankaios_server/state_comparator.rs
inf17101: I have done renamings.

PR #616, opened 2025-11-03
`ank get events` support

HorjuRares:
Issues: #450

- Introduced new command for retrieving events with specified output format and field mask.
- Implemented event subscription and listening in the server connection module.
- Updated Cargo.toml and Cargo.lock to include the chrono dependency.

# Definition of Done

### DIFF
--- a/tests/resources/ankaios.resource
+++ b/tests/resources/ankaios.resource
@@ -1099,10 +1099,10 @@ the event output shall contain desiredState
 the event output shall contain altered fields with removed workloads
     Event Output Shall Contain Altered Fields With Removed Workloads    target/robot_tests_result/${TEST_FOLDER_NAME}/events_output.txt
 
-all workloads of agent "${agent_names}" shall start via events
+all workloads of agent "${agent_names}" have left the initial execution state
     User Waits For All Workloads To Start Via Events    ${agent_names}    60
 
-all workloads of agent "${agent_names}" shall start via events within "${timeout}" seconds
+all workloads of agent "${agent_names}" have left the initial execution state within "${timeout}" seconds
     User Waits For All Workloads To Start Via Events    ${agent_names}    ${timeout}
 
 the workload "${workload_name}" shall reach state "${expected_state}" on agent "${agent_name}" via events
--- a/tests/stests/control_interface/control_interface.robot
+++ b/tests/stests/control_interface/control_interface.robot
@@ -31,7 +31,7 @@ Test Ankaios workload successful start-up without a Control Interface access
     # Preconditions
     Given Ankaios server is started with config "${CONFIGS_DIR}/simple.yaml"
     And Ankaios agent is started with name "${agent_name}"
-    And all workloads of agent "${agent_name}" shall start via events
+    And all workloads of agent "${agent_name}" have left the initial execution state
     # Actions
     # Asserts
     Then the mount point for the control interface has not been generated for ${agent_name}
@@ -44,7 +44,7 @@ Test Ankaios workload restart after update without a Control Interface access
     # Preconditions
     Given Ankaios server is started with config "${CONFIGS_DIR}/simple_with_control.yaml"
     And Ankaios agent is started with name "${agent_name}"
-    And all workloads of agent "${agent_name}" shall start via events
+    And all workloads of agent "${agent_name}" have left the initial execution state
     # Actions
     When user triggers "ank -k apply ${CONFIGS_DIR}/simple.yaml"
     # Asserts
@@ -58,7 +58,7 @@ Test Ankaios workload restart after update with a Control Interface access
     # Preconditions
     Given Ankaios server is started with config "${CONFIGS_DIR}/simple.yaml"
     And Ankaios agent is started with name "${agent_name}"
-    And all workloads of agent "${agent_name}" shall start via events
+    And all workloads of agent "${agent_name}" have left the initial execution state
     And the mount point for the control interface has not been generated for ${agent_name}
     # Actions
     When user triggers "ank apply ${CONFIGS_DIR}/simple_with_control.yaml"
@@ -73,7 +73,7 @@ Test Ankaios containerd workload restart after update with a Control Interface a
     # Preconditions
     Given Ankaios server is started with config "${CONFIGS_DIR}/simple_containerd.yaml"
     And Ankaios agent is started with name "${agent_name}"
-    And all workloads of agent "${agent_name}" shall start via events
+    And all workloads of agent "${agent_name}" have left the initial execution state
     And the mount point for the control interface has not been generated for ${agent_name}
     # Actions
     When user triggers "ank apply ${CONFIGS_DIR}/simple_with_control_containerd.yaml"
@@ -96,7 +96,7 @@ Test workload with empty Control Interface access field mask rejected
     # Preconditions
     Given Ankaios server is started with config "${CONFIGS_DIR}/simple.yaml"
     And Ankaios agent is started with name "${agent_name}"
-    And all workloads of agent "${agent_name}" shall start via events
+    And all workloads of agent "${agent_name}" have left the initial execution state
     # Actions
     When user triggers "ank apply ${CONFIGS_DIR}/faulty_with_control_as_empty.yaml"
     # Asserts
@@ -118,7 +118,7 @@ Test Ankaios podman-kube workload restart after update without a Control Interfa
     # Preconditions
     Given Ankaios server is started with config "${CONFIGS_DIR}/simple_kube_with_control.yaml"
     And Ankaios agent is started with name "${agent_name}"
-    And all workloads of agent "${agent_name}" shall start via events
+    And all workloads of agent "${agent_name}" have left the initial execution state
     And podman kube has assigned an id for pod "simple-pod" of workload "simple-kube" on agent "${agent_name}"
     # Actions
     When user triggers "ank apply ${CONFIGS_DIR}/simple_kube.yaml"
@@ -132,7 +132,7 @@ Test Ankaios podman-kube workload restart after update with a Control Interface 
     # Preconditions
     Given Ankaios server is started with config "${CONFIGS_DIR}/simple_kube.yaml"
     And Ankaios agent is started with name "${agent_name}"
-    And all workloads of agent "${agent_name}" shall start via events
+    And all workloads of agent "${agent_name}" have left the initial execution state
     And the mount point for the control interface has not been generated for ${agent_name}
     # Actions
     When user triggers "ank apply ${CONFIGS_DIR}/simple_kube_with_control.yaml"
@@ -147,7 +147,7 @@ Test target path from control interface access is limited to the designated pod 
     # Preconditions
     And Ankaios server is started with config "${CONFIGS_DIR}/multi_container_podman_kube.yaml"
     And Ankaios agent is started with name "${agent_name}"
-    And all workloads of agent "${agent_name}" shall start via events
+    And all workloads of agent "${agent_name}" have left the initial execution state
     And the workload "simple" shall have the execution state "Running(Ok)" on agent "${agent_name}"
     And the mount point for the control interface has been generated for ${agent_name}
     # Asserts
--- a/tests/stests/events/get_events.robot
+++ b/tests/stests/events/get_events.robot
@@ -56,7 +56,7 @@ Test Ankaios CLI get events receives multiple sequential events
     Given Podman has deleted all existing containers
     And Ankaios server is started with config "${CONFIGS_DIR}/simple.yaml"
     And Ankaios agent is started with name "agent_A"
-    And all workloads of agent "agent_A" shall start via events
+    And all workloads of agent "agent_A" have left the initial execution state
     # Actions
     When user starts event listener with format "json" and field mask "" in background
     And the user waits "3" seconds
@@ -93,7 +93,7 @@ Test Ankaios CLI get events with workload state changes
     Given Podman has deleted all existing containers
     And Ankaios server is started with config "${CONFIGS_DIR}/simple.yaml"
     And Ankaios agent is started with name "agent_A"
-    And all workloads of agent "agent_A" shall start via events
+    And all workloads of agent "agent_A" have left the initial execution state
     # Actions
     When user starts event listener with format "yaml" and field mask "workloadStates" in background
     And the user waits "2" seconds
@@ -114,7 +114,7 @@ Test Ankaios CLI get events includes timestamp in output
     Given Podman has deleted all existing containers
     And Ankaios server is started with config "${CONFIGS_DIR}/simple.yaml"
     And Ankaios agent is started with name "agent_A"
-    And all workloads of agent "agent_A" shall start via events
+    And all workloads of agent "agent_A" have left the initial execution state
     # Actions
     When user starts event listener with format "json" and field mask "" in background
     And the user waits "3" seconds
@@ -134,7 +134,7 @@ Test Ankaios CLI get events with empty field mask
     Given Podman has deleted all existing containers
     And Ankaios server is started with config "${CONFIGS_DIR}/default.yaml"
     And Ankaios agent is started with name "agent_A"
-    And all workloads of agent "agent_A" shall start via events
+    And all workloads of agent "agent_A" have left the initial execution state
     # Actions
     When user starts event listener with format "yaml" and field mask "" in background
     And the user waits "3" seconds
@@ -156,7 +156,7 @@ Test Ankaios CLI get events with workload deletion
     And Ankaios server is started with config "${CONFIGS_DIR}/default.yaml"
     And Ankaios agent is started with name "agent_A"
     And Ankaios agent is started with name "agent_B"
-    And all workloads of agent "agent_A and agent_B" shall start via events
+    And all workloads of agent "agent_A and agent_B" have left the initial execution state
     # Actions
     When user starts event listener with format "json" and field mask "desiredState.workloads" in background
     And the user waits "3" seconds
--- a/tests/stests/workloads/reject_state_with_cyclic_dependencies.robot
+++ b/tests/stests/workloads/reject_state_with_cyclic_dependencies.robot
@@ -57,7 +57,7 @@ Test Ankaios CLI update state with cycle in interworkload dependencies is reject
     And Ankaios server is started with config "${default_state_yaml_file}"
     And Ankaios agent is started with name "agent_B"
     And Ankaios agent is started with name "agent_A"
-    And all workloads of agent "agent_A and agent_B" shall start via events
+    And all workloads of agent "agent_A and agent_B" have left the initial execution state
     # Actions
     And user triggers "ank -k set state ${new_state_yaml_file} desiredState.workloads.workload_C"
     # Asserts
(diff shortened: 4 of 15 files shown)


HorjuRares:
@krucod3 I have a local integration branch were both `ank get events` and the event handler are merged into the feature branch and the system tests are passing without any problem. Please proceed further with the review of this PR.

inf17101:
@HorjuRares: You can still update the branch with the one from the event handler.

ank/doc/swdesign/README.md
```diff
@@ -1055,6 +1055,107 @@ Needs:
...
+- impl
+- utest
+- stest
+
+#### CLI handles event errors and interruptions
+`swdd~cli-handles-event-subscription-errors~1`
+
+Status: approved
+
+When the CLI is listening for events and encounters an error response from the server, a termination signal (SIGINT, SIGTERM, SIGHUP, or SIGQUIT), or a connection interruption, the CLI shall report the error condition and gracefully terminate.
```
inf17101: I think there is already a signal handling module in the cli. Can you check this again? If so, we do not need that detailed precondition with all the supported signals again.

GabyUnalaq: Looking at the implementation, the signal handling used is located in `ank/src/cli_signals.rs` and covers those 4 signals.
In addition, there is a requirement (`swdd~cli-provides-termination-signal-handling~1`) that documents this signals.

Because the signals are already documented, there is no need to write them again in the cli event requirement.

ank/src/cli_commands/get_events.rs
```diff
@@ -0,0 +1,520 @@
...
+use crate::cli_error::CliError;
+use crate::output;
+use crate::output_debug;
+
+use ankaios_api::ank_base::{AlteredFields, CompleteState, CompleteStateResponse};
+use chrono::Utc;
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Clone, Default, PartialEq, Serialize, Deserialize)]
+struct EventOutput {
```
inf17101, line 27: You need this rename below, otherwise all system tests are going into the timeout because they are trying to get `completeState` from the event dictionary but the key exists as snake case `complete_state`. I verified this also with the CLI with `-o json`.

```suggestion
#[serde(rename_all = "camelCase")]
struct EventOutput {
```

I see also general problems with the current implementation of the event usage in the stests:

1. The code is checking for `Pending(Initial)`, but currently, we do not send events after starting up the server with a manifest. So pending initial states are never received after a startup only after an update in a running cluster! However some stests are using startup manifests, so they are affected. => if we should send those events at startup will be discussed on Wednesday in the team.
2. The current sequential order of the keywords in the stests can cause timing issues in the event subscriptions of the CLI. Explanation: The current order is 1. Server startup, 2. agent startups, 3. CLI event subscriptions to wait for initial execution states => if the event subscription is delayed the stest might miss important events and runs into the timeout for checking initial states. The correct order is: 1. Server startup, 2. CLI event subscriptions, 3. agent startups
3. If the timeout is reached caused by empty dictionaries (`complete_state`, `workload_states`, `agent_workloads` in the python code), a warning is output, but the stests are still passing.
4. When for a lot of stests the timeout of 60 sec is hit to wait for initial execution state events, the whole CI/CD pipeline can take 50 min runtime like here in this run: https://github.com/eclipse-ankaios/ankaios/actions/runs/20713173138/job/59458188540?pr=616
5. If there is no startup manifest provided to the server always the timeout of 60 sec is hit leading to delays in the stests => in this case we can remove the wait for initial state via events because there are no workloads and a manifest is applied later with `ank apply`. I will mark an example test in the PR where this is the case.

Even after fixing the camel case issue, 37 stests are failing locally on my side. Let's discuss the current approach.

server/src/ankaios_server/event_handler.rs
```diff
@@ -11,6 +11,7 @@
 // under the License.
 //
 // SPDX-License-Identifier: Apache-2.0
+
```
inf17101, line 14: This can be removed.

ank/doc/swdesign/README.md
```diff
@@ -1055,6 +1055,107 @@ Needs:
...
+### `ank get events`
+
+![Get events](plantuml/seq_get_events.svg)
+
+#### CLI provides get events command
+`swdd~cli-provides-get-events-command~1`
+
+Status: approved
+
+The Ankaios CLI shall provide a function to get real-time state change events from the Ankaios server with options to specify output format (JSON or YAML) and field mask for filtering events to specific parts of the state.
```
inf17101:
```suggestion
The Ankaios CLI shall provide a function to get real-time state change events from the Ankaios server with the following options to specify the:
* output format (JSON or YAML)
* field mask for filtering events to specific parts of the state.
```

ank/doc/swdesign/README.md
```diff
@@ -1055,6 +1055,107 @@ Needs:
...
+
+#### CLI subscribes for events from server
+`swdd~cli-subscribes-for-events~1`
+
+Status: approved
+
+When the user invokes the CLI with a request to get events, the CLI shall subscribe to the Ankaios server using a `CompleteStateRequest` message with the `subscribe_for_events` flag set to true and an optional field mask to filter events to specific parts of the state.
+
+Tags:
+- ServerConnection
```
inf17101, line 1089: There is an impl and an utest linked in the CliCommands file as well:
```suggestion
- ServerConnection
- CliCommands
```

ank/doc/swdesign/README.md
```diff
@@ -1055,6 +1055,107 @@ Needs:
...
+
+Needs:
+- impl
+- utest
+- stest
+
+#### CLI receives and processes events
```
inf17101:
```suggestion
#### CLI processes received events
```

ank/doc/swdesign/README.md
```diff
@@ -1055,6 +1055,107 @@ Needs:
...
+`swdd~cli-receives-events~1`
+
+Status: approved
+
+When the CLI has subscribed for events, the CLI shall continuously receive state change events from the Ankaios server, skip the initial complete state response as it represents a baseline snapshot, process only messages matching the subscription request ID, and ignore all other messages.
```
inf17101:
```suggestion
When the CLI is subscribed to events, the CLI shall:
* continuously receive state change events from the Ankaios server
* skip the initial complete state response as it represents a baseline snapshot
* process only messages matching the subscription request ID with ignoring all other messages.
```

ank/doc/swdesign/README.md
```diff
@@ -1055,6 +1055,107 @@ Needs:
...
+- impl
+- utest
+- stest
+
+#### CLI handles event errors and interruptions
+`swdd~cli-handles-event-subscription-errors~1`
+
+Status: approved
+
+When the CLI is listening for events and encounters an error response from the server, a termination signal, or a connection interruption, the CLI shall report the error condition and gracefully terminate.
```
inf17101:
```suggestion
When the CLI listens for events and encounters an error response from the server, a termination signal, or a connection interruption,
the CLI shall gracefully terminate with an appropriate error message.
```

ank/src/cli_commands/get_events.rs
```diff
@@ -0,0 +1,520 @@
...
+            output_format
+        );
+
+        // [impl->swdd~cli-subscribes-for-events~1]
+        let mut subscription = self
+            .server_connection
+            .subscribe_and_listen_for_events(object_field_mask)
+            .await?;
+
+        while let Some(event) = self
```
inf17101, line 78:
```suggestion
        // [impl->swdd~cli-handles-event-subscription-errors~1]
        while let Some(event) = self
```

ank/doc/swdesign/README.md
```diff
@@ -1055,6 +1055,107 @@ Needs:
...
+- impl
+- utest
+- stest
+
+#### CLI outputs events with metadata
+`swdd~cli-outputs-events-with-timestamp~1`
+
+Status: approved
+
+When the CLI receives an event from the Ankaios server, the CLI shall output the event with an RFC3339 formatted timestamp, altered fields information (added, updated, and removed fields), and the complete state using the same state representation format as the `get state` command.
```
inf17101:
```suggestion
When the CLI receives an event from the Ankaios server, the CLI shall output the event with:
* an RFC3339 formatted timestamp
* altered fields information (added, updated, and removed fields)
* the complete state using the same state representation format as the `get state` command.
```

ank/doc/swdesign/README.md
```diff
@@ -1055,6 +1055,107 @@ Needs:
...
+
+#### CLI supports configurable event output formats
+`swdd~cli-supports-multiple-output-types-for-events~1`
+
+Status: approved
+
+When the CLI receives events from the Ankaios server, the CLI shall support presenting the events in both YAML (default) and JSON formats as specified by the output format command-line argument.
+
+Tags:
+- Cli
```
inf17101:
```suggestion
```

tests/stests/events/get_events.robot
```diff
@@ -0,0 +1,171 @@
...
+# [stest->swdd~cli-supports-multiple-output-types-for-events~1]
+Test Ankaios CLI get events with field mask filter
+    [Documentation]    Subscribe to events with field mask and verify filtered output
+    [Setup]    Setup Ankaios
+
+    # Preconditions
+    Given Podman has deleted all existing containers
+    And Ankaios server is started without config successfully
+    And Ankaios agent is started with name "agent_A"
+    And all workloads of agent "agent_A" shall start via events
```
inf17101: This line can be omitted, otherwise the line will block 60 sec and will not get an event because there are no events because of an empty manifest upon server startup. I think this is the only occurrence.

tests/resources/ankaios_library.py
```diff
@@ -858,3 +1066,264 @@ def get_instance_name_from_ankaios_workload_states(workload_states: str, workloa
...
+
+@err_logging_decorator
+def user_waits_for_all_workloads_to_start_via_events(agent_name: str, timeout: str="10"):
+    timeout_float = float(timeout)
+    result = wait_for_initial_execution_state_via_events(
+        agent_name=agent_name,
+        timeout=timeout_float
+    )
+
+    assert result is None, \
```
inf17101: It is required to assert for "not None". Otherwise some stests are failing.
```suggestion
    assert result is not None, \
```

tests/resources/ankaios_library.py
```diff
@@ -265,6 +266,210 @@ def get_pod_id_by_pod_name_from_podman(pod_name: str) -> str:
...
+        )
+
+        start_time = get_time_secs()
+        last_position = 0
+        first_event_received = False
+
+        while (get_time_secs() - start_time) < timeout:
+            if process.poll() is not None:
+                stderr = process.stderr.read()
+                logger.warn(f"Event listener process terminated: {stderr}")
```
inf17101: In general I would agree with the `break` here, but in our stests sometimes the CLI cannot connect to server because the server might not be ready and running its API, and same like with `get state` command with the polling in the stests the error `Could not connect to Ankaios server on 'http://127.0.0.1:25551'` might happen in a stest but still passes because the subsequent get state calls will succeed when the server will be ready. However, this can also happen when doing a `ank get events` instead, but with the break the `return None` below is executed and the stest will fail with `Event listener process terminated: error: Could not connect to Ankaios server on 'http://127.0.0.1:25551'`. What should be done is to reconnect the `ank get events` or making sure the server is up and running when doing the event subscription. Because, currently stest are sporadically failing because of this.

tests/stests/workloads/reject_state_with_cyclic_dependencies.robot
```diff
@@ -56,9 +56,9 @@ Test Ankaios CLI update state with cycle in interworkload dependencies is reject
     Given Podman has deleted all existing containers
     And Ankaios server is started with config "${default_state_yaml_file}"
     And Ankaios agent is started with name "agent_B"
-    And all workloads of agent "agent_B" have an initial execution state
+    And all workloads of agent "agent_B" shall start via events
```
inf17101: In this case changing the order has no effect. This is the only stest failing in the pipeline after applying fixes. The problem here is that the dependencies are distributed over two agents, so we should support multiple agents in one blocking "shall start via events" statements to be independent of the order, when having multiple agents.

We can switch to something like this similiar to workload names we already have multiple times in the stest code:
```text
all workloads of agent agent_A and agent_B and agent_C ... shall start via events
```
and then split it with:
`agent_names.replace(" and ", ", ").split(", ")`

ank/src/cli_commands/get_events.rs
GabyUnalaq: I believe this is fixed now.

server/src/ankaios_server/event_handler.rs
GabyUnalaq: I would leave it there, looking at all the other scripts, after the copyright header, there should be an empty line.

ank/doc/swdesign/README.md
GabyUnalaq: Looking at how other requirements are written; we can leave it as it was before. In addition, I believe we can remove the output format examples, because there is already another requirement covering the output formats that should be allowed by the ank cli (`swdd~cli-supports-multiple-output-types-for-events~1`)

ank/doc/swdesign/README.md
GabyUnalaq: Done.

ank/doc/swdesign/README.md
GabyUnalaq: Done.

ank/doc/swdesign/README.md
GabyUnalaq: Done.

ank/doc/swdesign/README.md
GabyUnalaq: Done.

ank/src/cli_commands/get_events.rs
GabyUnalaq: This requirement is linked already at the `receive_next_event` function. Should we link it here as well?

ank/doc/swdesign/README.md
GabyUnalaq: Done.

ank/doc/swdesign/README.md
GabyUnalaq: Done.

ank/src/cli_commands/get_events.rs
inf17101: No not all of those points.

server/src/ankaios_server/event_handler.rs
inf17101: ok let's keep it.

ank/src/cli_commands/get_events.rs
inf17101: For me it is strange, that the implementation is done in a lower level component with impl + utest, but the swdd is linked in utests in the upper level component. I think suggesting the minimal change here to just link the call also in the impl in the upper level module would be the easiest change for effort-time balance? Or you link another swdd from the higher level module to those tests

tests/stests/workloads/reject_state_with_cyclic_dependencies.robot
GabyUnalaq: Modified the current instruction to allow multiple agents. I allow only the `and` keyword to split so that the sentence sounds normal in the stest body.

ank/src/cli_commands/get_events.rs
GabyUnalaq: Commited now the fix.

ank/doc/swdesign/README.md
inf17101: This is fine, so removing the output format and then we do not need the enumeration.

tests/stests/events/get_events.robot
inf17101: Fixed already

tests/resources/ankaios_library.py
inf17101: Fixed by me

tests/stests/workloads/reject_state_with_cyclic_dependencies.robot
inf17101: Fixed in [3d82056](https://github.com/eclipse-ankaios/ankaios/pull/616/commits/3d82056dbda91f8cccc6fc763eaa597c6e829985)

ank/src/cli_commands/get_events.rs
inf17101: Fixed in [361d0fb](https://github.com/eclipse-ankaios/ankaios/pull/616/commits/361d0fb7b9e8f8a1f10248451a79d977fc105881)

tests/resources/ankaios.resource
```diff
@@ -1069,3 +1069,59 @@ the last command shall finish with an error
...
+the event output shall contain workloadStates
+    Event Output Shall Contain WorkloadStates    target/robot_tests_result/${TEST_FOLDER_NAME}/events_output.txt
+
+the event output shall contain desiredState
+    Event Output Shall Contain DesiredState    target/robot_tests_result/${TEST_FOLDER_NAME}/events_output.txt
+
+the event output shall contain altered fields with removed workloads
+    Event Output Shall Contain Altered Fields With Removed Workloads    target/robot_tests_result/${TEST_FOLDER_NAME}/events_output.txt
+
+all workloads of agent "${agent_names}" shall start via events
```
christoph-hamm: I think the "shall start via events" is misleading. This functions waits until everything is started. The names suggests it start somethings using events.

inf17101: Yes, good point. I talked with Gabi already about that and my plan is that if I anyway took over this branch, I will rename it after stabilizing all stests.

tests/stests/control_interface/control_interface.robot
```diff
@@ -58,7 +58,7 @@ Test Ankaios workload restart after update with a Control Interface access
     # Preconditions
     Given Ankaios server is started with config "${CONFIGS_DIR}/simple.yaml"
     And Ankaios agent is started with name "${agent_name}"
-    And all workloads of agent "${agent_name}" have an initial execution state
+    And all workloads of agent "${agent_name}" shall start via events
```
christoph-hamm: I am wonder if we could replace all usage of `wait_for_initial_execution_state` with `wait_for_initial_execution_state_via_events`.

tests/resources/ankaios_library.py
inf17101: fixed in [0543da1](https://github.com/eclipse-ankaios/ankaios/pull/616/commits/0543da13a095d217e36d84214fb745d8b766a69c).

tests/stests/control_interface/control_interface.robot
inf17101: Fixed in [a581303](https://github.com/eclipse-ankaios/ankaios/pull/616/commits/a5813035a48536e0f7a33a0b906b52ebcd7b9b68).

ank/doc/swdesign/README.md
inf17101: Fixed in [ac40af0](https://github.com/eclipse-ankaios/ankaios/pull/616/commits/ac40af00115a364b2939c0973e056cf08a4ad52a).

tests/resources/ankaios.resource
inf17101: Renamed in [0af5c9d](https://github.com/eclipse-ankaios/ankaios/pull/616/commits/0af5c9dd687f77295bd0cef9e522b3d5fb2815c4)

ank/src/cli_commands/get_events.rs
inf17101: The order of subscribing with events for stests and also replacing most of the polling variants is done in a separate PR to unblock the event features and this PR. I will commit my local stests changes to a separate branch and open a new PR.

inf17101: PR #663

PR #642, opened 2025-12-16
Treat empty subscriber mask as wildcard

inf17101:
Issues: #450 

Currently, an empty subscriber field mask results in no events. To make it similar to the `get state` interface, we now also support an empty subscriber mask which shall behave like a `*` (subscribe to everything).

# Definition of Done

### DIFF
--- a/server/doc/swdesign/README.md
+++ b/server/doc/swdesign/README.md
@@ -564,6 +564,7 @@ When the Ankaios Server responses to a GetCompleteState request and the request 
 the response includes the filed `api_version` and the fields listed in the `field_mask`.
 
 Comment:
+If the `field_mask` is empty, the response includes all fields of the CompleteState.
 If the fields listed in the `field_mask` contain templated strings, they are returned unrendered.
 
 Tags:


GabyUnalaq (approved):
LGTM 👍 

The only thing: should we add a requirement for the get state as well (probably on `server/src/ankaios_server/server_state.rs`, in `get_complete_state_by_field_mask`)? It is the same situation, if the field mask is empty, the result is the same as having it as a wildcard. I may be wrong but I don't see any requirement documenting the empty field mask situation, except now the event one.
There is a utest apparently `utest_server_state_get_complete_state_by_field_mask_empty_mask` for the get_state, but again, the requirement does not mention the empty field mask.

inf17101:
> 2386c17

Good point. I added it as a comment in [2386c17](https://github.com/eclipse-ankaios/ankaios/pull/642/commits/2386c17e437cb105b00c74a5208d1993bfb0fa22).

GabyUnalaq (approved):
LGTM 👍

PR #643, opened 2025-12-16
Fix deprecated proto extension

GabyUnalaq:
# Description

The proto3 extension that we were using is deprecated. This PR fixes this, by changing the extension.

With this extension, the proto3 linter works as expected (so I had to disable the naming conventions lint).

# Definition of Done

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -44,5 +44,9 @@
         "tag:yaml.org,2002:python/name:materialx.emoji.twemoji",
         "tag:yaml.org,2002:python/name:pymdownx.superfences.fence_code_format"
     ],
-    "plantuml.jar": "/usr/local/bin/plantuml.jar"
+    "plantuml.jar": "/usr/local/bin/plantuml.jar",
+    "protobuf.diagnostics.namingConventions": false,
+    "protobuf.includes": [
+        "/workspaces/ankaios/ankaios_api/proto"
+    ]
 }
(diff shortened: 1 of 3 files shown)


.vscode/settings.json
```diff
@@ -44,5 +44,9 @@
         "tag:yaml.org,2002:python/name:materialx.emoji.twemoji",
         "tag:yaml.org,2002:python/name:pymdownx.superfences.fence_code_format"
     ],
-    "plantuml.jar": "/usr/local/bin/plantuml.jar"
+    "plantuml.jar": "/usr/local/bin/plantuml.jar",
+    "protobuf.diagnostics.namingConventions": false,
+    "protobuf.includes": [
+        "/workspaces/ankaios/ankaios_api/proto"
+    ]
```
krucod3, line 51: I suppose this is only for the includes s.t. no errors are shown in grpc for example. Can we also specify a distinct file? This way nobody will use the control_api stuff in grpc.

krucod3: I just tried and it only accepts a folder.

krucod3 (approved):
LGTM 👍

.vscode/settings.json
GabyUnalaq: Because of the way we use the proto files in the grpc proto, we need to specify the folder of the proto files, not the proto itself, so it does not work. As for using control_api inside the grpc, we would need to import it at the top, which is not there at the moment, so it should be fine.

Issue #567, opened 2025-08-18
Configure runtimes to be enabled


windsource:
## Description

As of v0.7 Ankaios will support two runtimes: podman and containerd. In rare cases users might want to use both runtimes but usually a user sticks to one of them. So when a user only uses the containerd runtime, then the podman binary might not be installed and vice versa. 

When podman is started without the podman binary being installed, there are some WARN logs like:

```
[2025-08-18T08:54:55Z INFO  ank_agent] Loading agent config from default path '/etc/ankaios/ank-agent.conf'
[2025-08-18T08:54:55Z INFO  ank_agent::agent_manager] Awaiting commands from the server ...
[2025-08-18T08:54:55Z INFO  ank_agent::runtime_manager] Received the server hello with '0' added workloads.
[2025-08-18T08:54:55Z WARN  ank_agent::runtime_manager] Could not get reusable running workloads: 'Error: 'No such file or directory (os error 2)'. Could not spawn command 'Command { std: "podman" "ps" "--all" "--filter" "label=agent=agent_A" "--format=json", kill_on_drop: false }'.'
[2025-08-18T08:54:55Z WARN  ank_agent::runtime_manager] Could not get reusable running workloads: 'Could not list volume containing config: 'Error: 'No such file or directory (os error 2)'. Could not spawn command 'Command { std: "podman" "volume" "ls" "--filter" "name=.agent_A.config$" "--format={{.Name}}", kill_on_drop: false }'.''
```

But the agent can still be used with the other available runtimes.

To have that cleaned up it would be good to have a configuration option in the agent config file `ank-agent.conf` to state the runtimes that should be enabled at startup like:

```toml
runtimes = [ "podman", "containerd" ]
```

Only those runtimes should then be started by the agent. If the array is missing or empty then all runtimes should be started.

## Goals

Avoid warning messages when an Ankaios agent is started and binary for a runtime is missing that will never be used.

### Summary

**To be filled when the final solution is sketched.**

### Tasks

- [ ] Task 1
- [ ] Task 2
- [ ] ...

PR #646, opened 2025-12-22
Support for runtime whitelisting

GabyUnalaq:
Issues: #567

This PR adds a new config file field for enabling runtimes for the Ankaios agent:

```conf
runtimes = [ "podman", "containerd" ]
```

# Fields logic

The logic should be:
- If the field is not found, then a default list should be available, containing all of the available runtimes;
- If the field is found, the runtimes will be checked and the default list will be overwritten by the valid ones
  + if the list contains only invalid runtimes or if the list is empty => the agent is started without any runtimes => the agent should fail.

If a workload is started with a runtime which is not enabled, the behavior would be the same as an Unsupported Runtime.

# Definition of Done

### DIFF
--- a/agent/doc/swdesign/README.md
+++ b/agent/doc/swdesign/README.md
@@ -469,7 +469,7 @@ Needs:
 
 Status: approved
 
-The Ankaios agent shall support enabling a specific set of runtimes through the config file, with a fallback to all available runtimes if no valid ones were configured.
+The Ankaios agent shall support enabling a specific set of runtimes through the config file, with a fallback to all available runtimes if no whitelist is provided.
 
 Rationale:
 This allows the users to configure which runtimes are allowed on a specific agent, thus needing to make sure only the matching binaries are present.


agent/config/ank-agent.conf
```diff
@@ -21,6 +21,10 @@ name = 'agent_A'
 # If set to 'true', then the agent shall not use TLS.
 insecure = true
 
+# The list of enabled runtimes. If the array is missing
+# or empty, then all runtimes are allowed.
```
krucod3: I personally find the part with `empty` => all is enabled a bit strange. It's kind of a whitelist that allows all if empty.

@windsource, how about going for an explicit black list here with `disabled_runtimes`? I think it would be easier to understand this way.

agent/src/main.rs
```diff
@@ -112,6 +112,48 @@ pub fn validate_agent_name(agent_name: &str) -> Result<(), String> {
...
+                if SUPPORTED_RUNTIMES.contains(&runtime.as_str()) {
+                    valid_runtimes.push(runtime.as_str());
+                } else {
+                    log::warn!("Configured runtime '{runtime}' is not supported.");
+                }
+            }
+
+            if valid_runtimes.is_empty() {
+                log::warn!("No valid runtimes configured. Falling back to all supported runtimes.");
+                SUPPORTED_RUNTIMES.to_vec()
```
krucod3: So if I want to enable only podman, but write poodman instead, all other runtimes will be enabled too.

agent/config/ank-agent.conf
windsource: I would stick to the whitelist but agree that an empty list should exclude all runtimes. So I think we should have a default value of the whitelist hardcoded that includes all supported runtimes which can be overwritten by the config file.

GabyUnalaq: I have updated the description.

agent/src/main.rs
GabyUnalaq: Now by writing `poodman`, the output will be the following:

```bash
[2026-01-09T17:04:06Z INFO  ank_agent] Loading agent config from user provided path '/workspaces/ankaios/agent/config/ank-agent.conf'
[2026-01-09T17:04:06Z WARN  ank_agent] Configured runtime 'poodman' is not supported.
[2026-01-09T17:04:06Z ERROR] Invalid runtime configuration: 'No valid runtimes configured. Supported runtimes: ["podman", "containerd", "podman-kube"]'
```

agent/doc/swdesign/README.md
```diff
@@ -464,6 +464,23 @@ Needs:
 - impl
 - utest
 
+#### Agent allows only enabled runtimes
+`swdd~agent-allows-enabled-runtimes~1`
+
+Status: approved
+
+The Ankaios agent shall support enabling a specific set of runtimes through the config file, with a fallback to all available runtimes if no valid ones were configured.
```
krucod3:
```suggestion
The Ankaios agent shall support enabling a specific set of runtimes through the config file, with a fallback to all available runtimes if no whitelist is provided.
```

Issue #442, opened 2025-01-30
Flexible agent attributes (tags) in Complete State


krucod3:
## Description

The Complete State currently provides fixed agent resource statistics (CPU usage and free memory). 
Additional flexible agent tags can be added to the agent object which would help third-party orchestration plugins schedule workloads.

The idea is to allow updating certain attributes of the agents object (i.e. tags) both through the Control Interface and through the CLI. Additionally providing agent tags should also be possible at startup over cli arguments, env variables and config files.

## Goals

Enable scheduling plugins to distinguish between agents depending on special attributes of the nodes, e.g., 
* provided resources (like camera, temperature sensor, etc.), 
* node location (in-vehicle, cloud), 
* node connectivity (connected to cluster over ethernet or CAN)
* etc

## Final result

The following ways shall be available for adding/modifying/deleting tags:
* ank cli (Command Interface)
* Control Interface
* agent startup (over cli arguments, env variables and config files)

### Summary

**To be filled when the final solution is sketched.**

### Tasks

- [ ] Task 1
- [ ] Task 2
- [ ] ...

PR #649, opened 2025-12-22
Flexible agent attributes (tags) in Complete State

GabyUnalaq:
Issues: #442

This PR adds tags to an agent. TODO list:

- [x] Allow setting tags in the `ank-agent.conf` file;
- [x] Allow setting tags at the cli when starting the agent;
- [ ] ~Allow setting tags by using env variables~;
- [x] Send agent tags to the server to save in the CompleteState;
- [x] Allow changing tags through the Control Interface;
- [x] Allow changing tags through the use of the Ank CLI;
- [x] Test implementation using an stest;
- [x] Implement requirements;

# Definition of Done

### DIFF
--- a/agent/config/ank-agent.conf
+++ b/agent/config/ank-agent.conf
@@ -23,7 +23,7 @@ insecure = true
 
 # The list of enabled runtimes. If the array is missing
 # all runtimes will be enabled.
-runtimes = [ "podman", "containerd" ]
+# runtimes = [ "podman", "containerd" ]
 
 # Set of flexible agent attributes as key-value pairs.
 # tags = { "cpu" = "x86_64", "location" = "cloud", "camera" = "available" }


agent/config/ank-agent.conf
```diff
@@ -21,6 +21,10 @@ name = 'agent_A'
 # If set to 'true', then the agent shall not use TLS.
 insecure = true
 
+# Set of flexible agent attributes as key-value pairs.
+# Example: tags = { "cpu" = "x86_64", "location" = "cloud", "camera" = "available" }
+tags = {}
```
krucod3: We can list it commented out
```suggestion
# tags = { "cpu" = "x86_64", "location" = "cloud", "camera" = "available" }
```

agent/src/agent_config.rs
```diff
@@ -174,6 +178,12 @@ impl AgentConfig {
             let key_pem_content = read_pem_file(key_pem_path, true).unwrap_or_default();
             self.key_pem_content = Some(key_pem_content);
         }
+
+        if let Some(tags) = &args.tags {
+            for (key, value) in tags {
+                self.tags.insert(key.clone(), value.clone());
+            }
+        }
```
krucod3, line 187: We should overwrite here and not append. This is the behavior also for the other attributes.

agent/src/agent_config.rs
```diff
@@ -375,4 +393,74 @@ mod tests {
...
+        assert_eq!(agent_config.tags.get("cpu"), Some(&"x86_64".to_string())); // Overridden
+        assert_eq!(
+            agent_config.tags.get("camera"),
+            Some(&"available".to_string())
+        );
+        assert_eq!(
+            agent_config.tags.get("location"),
+            Some(&"cloud".to_string())
+        );
+    }
```
krucod3, line 426: As we are inserting, the tags from the config file are still available. We should update the code and add the check:
```suggestion
        assert!(
            agent_config.tags.get("camera").is_none()
        );
    }
```

Maybe we should also call the test "overwrite" instead of "override".

agent/src/cli.rs
```diff
@@ -50,6 +54,14 @@ pub struct Arguments {
     pub key_pem: Option<String>,
 }
 
+/// Parse a single key-value pair
+fn parse_key_val(s: &str) -> Result<(String, String), String> {
+    let pos = s
+        .find('=')
+        .ok_or_else(|| format!("invalid KEY=value: no `=` found in `{s}`"))?;
```
krucod3: The error in the CLI now looks like this:
```
error: invalid value 'testtag' for '--tag <TAGS>': invalid KEY=value: no `=` found in `testtag`
```
Please change to:
```suggestion
        .ok_or_else(|| format!("no `=` found in '{s}', expected the format `key=value`"))?;
```

krucod3: I just saw that we already have a `parse_key_val` function in the ank/src/cli.rs.

Please check and unify if it makes sense.

server/src/ankaios_server/server_state.rs
```diff
@@ -341,8 +342,15 @@ impl ServerState {
         })
     }
 
-    fn set_desired_state(&mut self, new_desired_state: StateSpec) {
-        self.state.desired_state = new_desired_state;
+    fn set_complete_state(&mut self, new_complete_state: CompleteStateSpec) {
```
krucod3: The name now suggests that attributes of the complete state are updated, which is not the case.

I would suggest to update the agent tags in a dedicated function.

krucod3 (commented):
Reviewed, findings documented.

agent/config/ank-agent.conf
GabyUnalaq: Changed it.

agent/src/cli.rs
GabyUnalaq: I moved the ank function to common and I used it in the agent as well.

server/src/ankaios_server/server_state.rs
GabyUnalaq: The function before contained just one line, which was a bit anticlimactic, and now it is used to update all the (update-able) fields of ServerState.state (which is a CompleteStateSpec). Another solution would be to rename the function to something more appropriate, to not give the false premise that the whole of the complete state gets updated.

krucod3: Well ... the old function was called `set_desired_state` and the name was fitting although the functionality was indeed a bit disappointing. Especially with the heavy usage of ... (drums increasing the tension) ... 2 spots only.
Still, the name was fitting. 

My suggestion was to leave the old function as as and update the tags in a dedicated one called update tags. If you want, you can delete the old function `set_desired_state` and replace with its internal code.

agent/src/agent_config.rs
GabyUnalaq: Fixed.

agent/src/agent_config.rs
GabyUnalaq: Due to the change in behavior, one utest was deleted. It should be fine now.

server/src/ankaios_server/server_state.rs
GabyUnalaq: Created a new "set_agent_tags" method for the agent tags.

krucod3 (approved):
LGTM 👍

Issue #396, opened 2024-10-21
Re-use existing container bundles for restart policy


windsource:
## Description

With PR #386 we re-use container bundles on device restart. But there is also another situation when a workload exits and a `restartPolicy` is provided and thus the workload is restarted. In that case when the workload has exited with state `SUCCEEDED` then we should not delete the existing container bundle but re-use it.

## Goals

Re-use container bundles when the restart policy is applied and workload is in state `SUCCEEDED`.

### Summary

**To be filled when the final solution is sketched.**

### Tasks

- [ ] Task 1
- [ ] Task 2
- [ ] ...

PR #650, opened 2025-12-29
Re-use existing container bundles when restarting succeeded workloads 

GabyUnalaq:
Issues: #396 

If a workload has the restart policy set to `ALWAYS` and it's in the `SUCCEEDED` state, then the workload should be restarted and the bundle reused.

- [x] Implement bundle reuse for succeeded case;
- [x] Fix and add a utests for checking this case;
- [x] Add requirement
- [x] Add stest

# Definition of Done

### DIFF
--- a/agent/doc/swdesign/README.md
+++ b/agent/doc/swdesign/README.md
@@ -1818,7 +1818,7 @@ Needs:
 
 Status: approved
 
-When the WorkloadControlLoop restarts a workload that has reached the ExecutionState `Succeeded`, then the WorkloadControlLoop shall reuse the existing bundle and workload files.
+When the WorkloadControlLoop executes a restart policy for a workload that has reached the ExecutionState `Succeeded`, then the WorkloadControlLoop shall reuse the existing bundle and workload files.
 
 Rationale:
 Reusing the bundle and workload files for successfully completed workloads improves restart performance by eliminating unnecessary deletion and recreation of resources that are already in the correct state. This optimization reduces I/O operations and restart time while maintaining system consistency.


agent/src/workload/workload_control_loop.rs
```diff
@@ -199,12 +201,21 @@ impl WorkloadControlLoop {
         let workload_named = control_loop_state.workload_named.clone();
         let control_interface_path = control_loop_state.control_interface_path.clone();
 
+        // Check if we can reuse the bundle: workload succeeded and has a workload_id
+        let reusable_workload_id =
+            if execution_state.is_succeeded() && control_loop_state.workload_id.is_some() {
+                control_loop_state.workload_id.clone()
```
krucod3: We clone here, but later only use a flag.

GabyUnalaq: I've refactored a bit and now I only send a boolean. I've also prepared the code in such a way that there is only one place to add the requirement, once it's added.

GabyUnalaq: After a discussion, I have split it to another function. I believe now it should be better than ever

agent/doc/swdesign/README.md
```diff
@@ -1813,6 +1813,24 @@ Needs:
 - impl
 - utest
 
+#### WorkloadControlLoop reuses bundle on successful workload restart
+`swdd~workload-control-loop-reuses-bundle-on-successful-restart~1`
+
+Status: approved
+
+When the WorkloadControlLoop restarts a workload that has reached the ExecutionState `Succeeded`, then the WorkloadControlLoop shall reuse the existing bundle and workload files.
```
krucod3: I'm lacking a bit the reason for the restart:
```suggestion
When the WorkloadControlLoop executes a restart policy for a workload that has reached the ExecutionState `Succeeded`, then the WorkloadControlLoop shall reuse the existing bundle and workload files.
```

Issue #264, opened 2024-05-02
Replace the polling in system tests


inf17101:
## Description

Currently, the stests are implemented to use polling to ask for workload states, workloads ,etc. This leads sometimes to unstable system tests and forces a developer to search for alternative solutions to make the stest stable. Sometimes such workarounds leads to weaker tests and it costs time to implement alternative solutions afterwards.

## Goals

Replace the polling method through a better mechanism to avoid missing information for assertions inside stests.

### Summary

**To be filled when the final solution is sketched.**

### Tasks

- [ ] Task 1
- [ ] Task 2
- [ ] ...

krucod3:
Depends on #228

inf17101:
Like discussed, even if we replace the CLI polling, internally the state checker still does polling every 1 sec. However, I have an suggestion for getting rid completely of the internal polling for podman and podman-kube. Podman supports `podman-events` CLI command that blocks and informs you about every event that occurs for a container or pod. Here is an example recording of me demonstrating `podman-events`:

https://github.com/user-attachments/assets/795609d8-7f18-4f75-b01a-5330d4a2d53a

This can be also considered to replace the internal polling as well (maybe we can also get rid of the cache). We can parse the lines podman events is reporting or add the `--filter=container=<container_name_or_id>` to the command to just get the events for a specific container.

I have checked the [podman-events documentation](https://docs.podman.io/en/stable/markdown/podman-events.1.html) and it seems like that this command is available also for previous podman versions (v3.0, ...).

PR #663, opened 2026-01-14
Replace polling in stests with events support

inf17101:
Issues: #264 #450 

Replaces stest polling a separate PR for easier review.

When the PR #616 is merged the base branch for this PR must be changed to the events feature branch. For now I want to test the stests with the event changes in the CI/CD as well.

# Definition of Done

### DIFF
--- a/tests/stests/mtls/mtls_communication.robot
+++ b/tests/stests/mtls/mtls_communication.robot
@@ -81,7 +81,7 @@ Test Ankaios MTLS by providing wrong PEM config via command line arguments
 # [stest->swdd~cli-supports-cli-argument-for-insecure-communication~1]
 # [stest->swdd~cli-establishes-insecure-communication-based-on-provided-insecure-cli-argument~1]
 Test Ankaios insecure mode by providing --insecure command line arguments
-    [Setup]    Run Keyword    Setup Ankaios Without MTLS Setup
+    [Setup]    Run Keyword    Setup Ankaios
     # Preconditions
     # This test assumes that all containers in the podman have been created with this test -> clean it up first
     Given Podman has deleted all existing containers


tests/resources/ankaios.resource
```diff
@@ -363,52 +345,49 @@ user updates the state "${new_state_yaml_file}" with "${field_value}"
...
+            IF    ${elapsed_secs} >= ${timeout}    BREAK
+
+            ${command_result}=    Run Command    %{ANK_BIN_DIR}${/}ank get workloads
+            @{list_result}=    table to list    ${command_result.stdout}
+            &{workload_table}=    table to dict    ${list_result}    WORKLOAD NAME
+            ${workload_entry}=    Get From Dictionary    ${workload_table}    ${workload_name}
+            ${execution_state}=    Get From Dictionary    ${workload_entry}    EXECUTION STATE
+            IF    '${execution_state}' == '${expected_execution_state}'    RETURN
+
+        END
```
krucod3, line 377: Please check the code here again. I don't see the fail when breaking the loop and not reaching the desired state. I suppose we should have a `Fail` after the loop in case we did not find what we expected. The loop also performs a busy waiting as there is not sleep in each iteration.

Additionally, to test if it would catch a different state, I changed the `Test Ankaios CLI lists connected agents` test to start an additional agent_B and it did not fail. Please have a look at the test and consider after fixing if we need the added complexity for the CLI polling.

inf17101: Fixed in [c35d7fb](https://github.com/eclipse-ankaios/ankaios/pull/663/commits/c35d7fba3a3cef206822ea43ff1931eefb9064bb) and I found also an additional assertion issue, which I fixed in [474451d](https://github.com/eclipse-ankaios/ankaios/pull/663/commits/474451d2e7537b1c1f2ab10503e6d450412f90e9), too.

inf17101: For this stest we can simplify the test to not use the `Pending(Initial)` expectation and we can remove the polling part. However, then this keyword mapping is not usable for every kind of execution state the user might input. For the initial pending state we need polling as it would not receive an event for initial states. I would leave the polling as the check for pending initial is fast. Very less amount of loop iterations would be required in the good case.

tests/resources/ankaios_library.py
```diff
@@ -1289,48 +1013,203 @@ def event_output_shall_contain_altered_fields_with_removed_workloads(output_file
...
+    global EVENT_PROCESS
+    logger.trace(f"Process is running: {EVENT_PROCESS}")
     timeout_float = float(timeout)
-    agent_names_split = [agent_names.strip() for agent_names in agent_names.split('and')]
-    logger.trace(f"Waiting for all workloads on agents {agent_names_split} to leave Pending(Initial) state")
-    result = wait_for_initial_execution_state_via_events(
-        agent_names=agent_names_split,
-        timeout=timeout_float
-    )
+    agent_names = { agent_name.strip(): False for agent_name in agent_names.split('and') }
```
inf17101: We need the initial list of workloads here, otherwise it might miss some states and returns too early!

inf17101: I implemented a solution local, but let's discuss it tomorrow.

inf17101: I removed the `initial workload state assertion` to simplify the keyword mappings and the stests. I went through all the stests and switched to `shall have the execution state...` keyword directly.

tests/stests/configs/config_file.robot
```diff
@@ -46,10 +46,10 @@ Test agent config file successful start-up
...
-    And Ankaios server is available
-    And Ankaios insecure agent is started with name "agent_A"
-    And all workloads of agent "agent_A" have left the initial execution state
+    Ankaios server is started with manifest "${CONFIGS_DIR}/default.yaml" and server config file "${CONFIGS_DIR}/ank-server.conf"
+    And the CLI listens for workload states
+    And Ankaios agent is started with name "agent_A"
+    And the workload "sleepy" shall have the execution state "Running(Ok)" on agent "agent_A"
     # Asserts
-    Then the workload "simple" shall not exist on agent "agent_A" within "1" seconds
+    Then the workload "simple" shall never be created on agent "agent_A" within "1" seconds
```
krucod3, line 54: Why not using the "the workload "${workload_name}" shall not exist" keyword and deleting this one?

tests/stests/mtls/mtls_communication.robot
```diff
@@ -79,21 +81,23 @@ Test Ankaios MTLS by providing wrong PEM config via command line arguments
 # [stest->swdd~cli-supports-cli-argument-for-insecure-communication~1]
 # [stest->swdd~cli-establishes-insecure-communication-based-on-provided-insecure-cli-argument~1]
 Test Ankaios insecure mode by providing --insecure command line arguments
-    [Setup]    Run Keyword    Setup Ankaios without MTLS Setup
+    [Setup]    Run Keyword    Setup Ankaios Without MTLS Setup
```
krucod3:
```suggestion
    [Setup]    Run Keyword    Setup Ankaios
```

tests/resources/ankaios.resource
```diff
@@ -324,6 +305,7 @@ Ankaios agent is started with name "${name}" and tags "${tags}"
     ...    alias=${name}
 
 Ankaios server is terminated
+    Unsubscribe From Events
```
krucod3: This implicit unsubscribe is a bit funny. It would also be clearer in the test that uses the `Ankaios server is terminated` (`Test Ankaios restarts exited workloads on device restart with considering inter-workload dependencies`) if the action `Unsubscribe From Events` is called explicitly.

tests/stests/configs/config_file.robot
inf17101: I have deleted this keyword mapping and only `shall never be created` is used now. It was redundant.

tests/resources/ankaios.resource
inf17101: Done in [96293e2](https://github.com/eclipse-ankaios/ankaios/pull/663/commits/96293e225e82c72dd030f9ca4d0a14cc005cb5a7)

Issue #414, opened 2024-11-12
Enhance `ank set state` functionality


HorjuRares:
### Summary

**To be filled when the final solution is sketched.**

### Tasks

- [ ] Task 1
- [ ] Task 2
- [ ] ...

krucod3:
This requires a fix the Ankaios server and agent as they must always accept incomplete objects when ToServer requests are handled.
@inf17101 even found the problem while going through eclipse-ankaios/ank-sdk-python#106.

We can fix the issues with this PR and test again both the SDK for correct behavior.

PR #669, opened 2026-01-22
ToServer Request must allow filtered data

krucod3:
Issues: #414

# Definition of Done

### DIFF
--- a/justfile
+++ b/justfile
@@ -60,7 +60,8 @@ utest:
     RUST_LOG=debug cargo nextest --config-file nextest.toml run
 
 # Build debug and run all system tests
-stest: build build-stest-image stest-only
+stest filter="*" tests="tests": build build-stest-image
+    just stest-only "{{ filter }}" "{{ tests }}"
 
 build-stest-image:
     #!/usr/bin/env bash
@@ -71,13 +72,9 @@ build-stest-image:
         echo 'Had to build control_interface_tester image. Consider uploading it with `podman push ghcr.io/eclipse-ankaios/control_interface_tester:'"$SRC_HASH"'`'
     fi
 
-# Build and run only the specified system test(s) (wildcard supported)
-stests-filter filter: build build-stest-image
-    ./tools/run_robot_tests.sh --test "{{ filter }}" tests
-
 # Only execute the stests without building
-stest-only tests="tests":
-    ./tools/run_robot_tests.sh {{ tests }}
+stest-only filter="*" tests="tests":
+    ./tools/run_robot_tests.sh --test "{{ filter }}" "{{ tests }}"
 
 # Run clippy code checks
 clippy:


agent/src/control_interface/control_interface_task.rs
```diff
@@ -286,11 +286,11 @@ mod tests {
     };
     use crate::test_helper::MOCKALL_CONTEXT_SYNC;
 
-    use ankaios_api::test_utils::fixtures;
+    use ankaios_api::{ank_base::LogsCancelRequest, test_utils::fixtures};
```
GabyUnalaq:
```suggestion
    use ankaios_api::test_utils::fixtures;
```
Move the `LogsCancelRequest` import with the others 3 lines bellow.

agent/src/control_interface/control_interface_task.rs
```diff
@@ -286,11 +286,11 @@ mod tests {
...
     use ankaios_api::{
         ank_base::{
-            CompleteStateRequest, Error as AnkError, LogEntriesResponse, LogsCancelRequestSpec,
-            Request, RequestContent, RequestContentSpec, RequestSpec, Response, ResponseContent,
+            CompleteStateRequest, Error as AnkError, LogEntriesResponse,
+            Request, RequestContent, Response, ResponseContent,
```
GabyUnalaq:
```suggestion
            CompleteStateRequest, Error as AnkError, LogEntriesResponse, LogsCancelRequest,
            Request, RequestContent, Response, ResponseContent,
```
Like so.

ank/src/cli_commands/server_connection.rs
```diff
@@ -783,42 +781,47 @@ mod tests {
...
+            id: fixtures::WORKLOAD_IDS[0].to_string(),
+        }
+    }
+
+    fn instance_name_spec(workload_name: &str) -> WorkloadInstanceNameSpec {
+        WorkloadInstanceNameSpec {
+            agent_name: fixtures::AGENT_NAMES[0].to_string(),
+            workload_name: workload_name.to_string(),
+            id: fixtures::WORKLOAD_IDS[0].to_string(),
+        }
```
GabyUnalaq, line 816: This functions are now duplicated code. Wouldn't be better to have only the spec and use it everywhere with type annotations and `.into()` called onto them?
Or maybe have both functions, but like:
```
fn instance_name(workload_name: &str) -> WorkloadInstanceName {
    instance_name_spec(workload_name).into()
}
```

justfile
```diff
@@ -71,6 +71,10 @@ build-stest-image:
         echo 'Had to build control_interface_tester image. Consider uploading it with `podman push ghcr.io/eclipse-ankaios/control_interface_tester:'"$SRC_HASH"'`'
     fi
 
+# Build and run only the specified system test(s) (wildcard supported)
+stests-filter filter: build build-stest-image
+    ./tools/run_robot_tests.sh --test "{{ filter }}" tests
+
 # Only execute the stests without building
 stest-only tests="tests":
     ./tools/run_robot_tests.sh {{ tests }}
```
GabyUnalaq: Because now we have 2 commands for stests, wouldn't it be better to merge them?
I've tested this and it works as expected:
```
stest-exec tests="tests" filter="*":
    ./tools/run_robot_tests.sh --test "{{ filter }}" {{ tests }}
```

krucod3: I wanted to prevent writing two arguments when executing two test, but have a very good point there. I'll `just` reorder the arguments 🤣

ank/src/cli_commands/server_connection.rs
krucod3: These are specific test fixture functions and I tried to optimize speed instead of clean code. given the low complexity, I prefer to leave them as they are now.

GabyUnalaq: Alright then.

justfile
GabyUnalaq: When executing, if you don't write arguments, the default ones will be used, so it's perfect

krucod3: Yes, of course, but the order matters too 😉

PR #674, opened 2026-01-23
Add cli flag for initial state response output in events

inf17101:
Issues: #450 

The PR introduces the following changes:

- `--include-current-state` (`-c`) event CLI flag for the cli to enable the output of the initial state response when subscribing to events
- removes the boolean flags in the `receive_next_event` function
- removes duplicated code for subscribing to the events with the CLI (now: reuse of `get_complete_state`)
- removes unsafe code in event unit tests
- refactors unit tests to also assert not only that events are received but also that event output is called
- changed an existing swdd to include the new cli detailed flag
- adds an stest for the new cli flag
- renaming of an stest keyword to subscribe to events

# Definition of Done

### DIFF
--- a/ank/src/cli_commands/get_events.rs
+++ b/ank/src/cli_commands/get_events.rs
@@ -577,7 +577,7 @@ mod tests {
 
     // [utest->swdd~cli-provides-get-events-command~2]
     #[tokio::test]
-    async fn utest_get_events_output_initial_state_response() {
+    async fn utest_get_events_output_current_state_response() {
         let _guard = crate::test_helper::MOCKALL_CONTEXT_SYNC
             .get_lock_async()
             .await;


ank/doc/swdesign/README.md
```diff
@@ -1064,7 +1064,11 @@ Needs:
 
 Status: approved
 
-The Ankaios CLI shall provide a function to get real-time state change events from the Ankaios server with a field mask as argument for filtering events to specific parts of the state.
+The Ankaios CLI shall provide a function to get real-time state change events from the Ankaios server with the following arguments:
```
inf17101, line 1067: Shall we increase the swdd id of this requirement?

ank/src/cli.rs
```diff
@@ -256,6 +256,9 @@ pub enum GetCommands {
         /// Specify the output format
         #[arg(short = 'o', value_enum, default_value_t = OutputFormat::Yaml)]
         output_format: OutputFormat,
+        /// Specify if the first complete state response shall be included in the output
+        #[arg(short = 'd', long = "detailed", default_value_t = false)]
```
inf17101: We can discuss about a better argument name with pleasure. I tried to avoid introducing a flag for already known intention.

tests/stests/events/get_events.robot
```diff
@@ -44,6 +44,24 @@ Test Ankaios CLI get events with field mask filter
...
+    And Ankaios server is available
+    And Ankaios agent is started with name "agent_A"
+    # Actions
+    When user starts the CLI to subscribe to events with detailed mode enabled with format "yaml" and field mask "desiredState.workloads" in background
+    And the user waits "1" seconds
+    # Asserts
+    Then the event output shall contain workload "nginx"
+    And the event output shall be valid yaml format
+    And the event output shall contain only desiredState workloads
+    [Teardown]    Clean up Ankaios
```
christoph-hamm: I think it would be better if this test is more similar to "Test Ankaios CLI get events with field mask filter", updating the state after subscribing to the events. Then we could test if getting the initial state and getting events work on the same time.

ank/doc/swdesign/README.md
christoph-hamm: As we not only made the requirement more specific, but changed how we understand it (the part with the details is added), I would increase the SwDD ID.

ank/src/cli.rs
christoph-hamm: Maybe "--include-current" and "-c"? I think it describes the argument better. The name "detailed" implies it return more information all the time, not only at the beginning.

inf17101: It is renamed to `--include-current-state` in [1196036](https://github.com/eclipse-ankaios/ankaios/pull/674/commits/11960364c951380d78289c69e01dc010cfb4dcbb)

tests/stests/events/get_events.robot
inf17101: I think this cannot be tested reliable, because the initial state response looks like an event in the CLI output and we cannot assert that it is the current state or the next event output when receiving something. Even counting number of received event responses is not a good choice. The current test setup is easy because you would not receive further events with this field mask, because nothing is changed in the state and the initial state response is the only one that is received and asserted. So if this misses the test would fail. It also tests exactly one thing, that the current state is output after subscribing.

ank/doc/swdesign/README.md
inf17101: The id is updated and the flag name is changed to the new one: [fb443d5](https://github.com/eclipse-ankaios/ankaios/pull/674/commits/fb443d5d215460c8e8105d6e0e595ff696d58f0e)

christoph-hamm: :+1:

tests/stests/events/get_events.robot
christoph-hamm: OK

Issue #512, opened 2025-07-03
Limit message size for workload logs


krucod3:
## Description

As there is a limitation on the gRPC message size of 4MiB, we should also take care that `LogEntriesResponse` messages do not exceed this limit.
Probably a lower limit is more reasonable.

This enhancement is extracted from the general ticket on workload logs #90

### Summary

**To be filled when the final solution is sketched.**

### Tasks

- [ ] Task 1
- [ ] Task 2
- [ ] ...

PR #675, opened 2026-01-27
Limit the size of logs

GabyUnalaq:
Issues: #512 

Logs should be processed in the following way:
* if a single log line exceeds 50MB during collection, an info message should be sent and then the log collection shall be interrupted
* if a single log entry exceeds 2MB when sending to the server, the log message shall be truncated and an info message should follow
* if multiple log entries together exceed 2MB, they shall be split into separate responses

# Definition of Done

### DIFF
--- a/agent/src/runtime_connectors/log_fetching/generic_log_fetcher.rs
+++ b/agent/src/runtime_connectors/log_fetching/generic_log_fetcher.rs
@@ -109,10 +109,11 @@ impl<T: AsyncRead + std::fmt::Debug + std::marker::Unpin> GenericSingleLogFetche
 
     // [impl->swdd~agent-checks-size-of-logs~1]
     async fn next_lines(&mut self) -> Option<Vec<String>> {
-        let mut start_byte = self.read_data.len();
         if self.limit_exceeded {
             return None;
         }
+
+        let mut start_byte = self.read_data.len();
         match self.reader.read_buf(&mut self.read_data).await {
             Ok(0) => {
                 if start_byte == 0 {
--- a/common/src/message_size.rs
+++ b/common/src/message_size.rs
@@ -99,7 +99,7 @@ fn truncate_log_entry(entry: LogEntry) -> (LogEntry, LogEntry) {
         }
         format!("{}{}", &entry.message[..end], TRUNCATION_NOTICE)
     } else {
-        entry.message.clone()
+        entry.message
     };
 
     let truncated_entry = LogEntry {
@@ -126,7 +126,7 @@ mod tests {
 
     // [utest->swdd~agent-checks-size-of-logs~1]
 
-    fn create_log_entry(message: &str) -> LogEntry {
+    fn generate_test_log_entry(message: &str) -> LogEntry {
         LogEntry {
             workload_name: Some(WorkloadInstanceName {
                 agent_name: fixtures::AGENT_NAMES[0].into(),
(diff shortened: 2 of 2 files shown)


inf17101:
I am starting to review this.

agent/src/runtime_connectors/log_fetching/generic_log_fetcher.rs
```diff
@@ -100,12 +102,17 @@ impl<T: AsyncRead + std::fmt::Debug + std::marker::Unpin> GenericSingleLogFetche
...
             read_data: BytesMut::new(),
         }
     }
 
+    // [impl->swdd~agent-checks-size-of-logs~1]
     async fn next_lines(&mut self) -> Option<Vec<String>> {
         let mut start_byte = self.read_data.len();
+        if self.limit_exceeded {
+            return None;
+        }
```
inf17101, line 114: First I was wondering where start_byte is initialzed in the code below. But then I saw it at the top before the early exit. Can we just switch this. Then it fits better to the "early-exit" style.

```suggestion
        if self.limit_exceeded {
            return None;
        }
        
        let mut start_byte = self.read_data.len();
```

agent/src/runtime_connectors/log_fetching/generic_log_fetcher.rs
```diff
@@ -118,7 +125,18 @@ impl<T: AsyncRead + std::fmt::Debug + std::marker::Unpin> GenericSingleLogFetche
...
+            Ok(_) => {
+                if self.read_data.len() > LOG_SIZE_LIMIT_MB * 1024 * 1024 {
+                    log::warn!(
+                        "Log line buffer exceeded {LOG_SIZE_LIMIT_MB} MB, closing the log campaign."
+                    );
+                    self.read_data.clear();
+                    self.limit_exceeded = true;
+                    return Some(vec![format!(
+                        "The previous log message exceeded the maximum {LOG_SIZE_LIMIT_MB} MB size. Closing the log campaign."
+                    )]);
```
inf17101: Here we mix the log output of applications with logs from Ankaios. For this we have the agent logs. We can maybe return an empty vector as a good state.

```suggestion
                    return Some(Vec::default());
```

agent/doc/swdesign/README.md
```diff
@@ -4313,6 +4313,23 @@ Needs:
...
 - utest
 
+#### Logs sizes should be checked
+`swdd~agent-checks-size-of-logs~1`
+
+Status: approved
+
+Logs should be processed in the following way:
+* if a single log line exceeds 50MB during collection, the log collection shall be interrupted
+* if a single log entry exceeds 2MB when sending to the server, the log message shall be truncated
```
inf17101, line 4323:
```suggestion
* if a single log response message exceeds 2MB, the log message shall be truncated
```

common/src/message_size.rs
```diff
@@ -0,0 +1,321 @@
...
+
+    // Truncate at char boundary
+    let truncated_message = if max_message_size < message_len {
+        let mut end = max_message_size;
+        while end > 0 && !entry.message.is_char_boundary(end) {
+            end -= 1;
+        }
+        format!("{}{}", &entry.message[..end], TRUNCATION_NOTICE)
+    } else {
+        entry.message.clone()
```
inf17101:
```suggestion
        entry.message
```

inf17101 (commented):
First half of the review.

common/src/message_size.rs
```diff
@@ -0,0 +1,321 @@
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use ankaios_api::ank_base::WorkloadInstanceName;
+    use ankaios_api::test_utils::fixtures;
+
+    // [utest->swdd~agent-checks-size-of-logs~1]
+
+    fn create_log_entry(message: &str) -> LogEntry {
```
inf17101: Just for agreed convention:

```suggestion
    fn generate_test_log_entry(message: &str) -> LogEntry {
```

common/src/message_size.rs
```diff
@@ -0,0 +1,321 @@
...
+            workload_name: Some(WorkloadInstanceName {
+                agent_name: fixtures::AGENT_NAMES[0].into(),
+                workload_name: fixtures::WORKLOAD_NAMES[0].into(),
+                id: fixtures::WORKLOAD_IDS[0].into(),
+            }),
+            message: message.into(),
+        }
+    }
+
+    fn create_log_entry_with_size(target_size: usize) -> LogEntry {
```
inf17101:
```suggestion
    fn generate_test_log_entry_with_size(target_size: usize) -> LogEntry {
```

inf17101 (commented):
Finished the review. Some small findings only.

agent/doc/swdesign/README.md
GabyUnalaq: There are 2 concepts at play: responses and entries
One response can have multiple entries, and the condition here targets the individual entry. Having a response here might not be the best.
I can write "if a single log message exceeds 2MB", this way it can refer to the message itself.

inf17101: Then I think the initial text was better. Let's keep it. Because with "log message" it is more weak than before.

PR #676, opened 2026-01-27
Configure VsCode Testing feature

GabyUnalaq:
# Description

This PR should allow the robot framework which we have in the devcontainer (`d-biehl.robotcode`) to work with the image Testing feature of the VSCode. This means that we will be able to run the system tests from that interface, easier than with the prior solution, using the script.

The basic changes include the extraction of the stests setup to another script, that is called in the init of the stests. This way, the setup will always be executed before running anything, redirecting the output to the stest normal output folder.

The simplification can go even further, if we get rid of the `run_robot_tests.sh` script (it became mostly one line) but the problem is the pipeline which still uses it. Opinions over this topic are welcomed.

# Definition of Done

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -55,6 +55,8 @@
     },
     "robotcode.robot.args": [
         "--pythonpath",
-        "tests"
+        "tests",
+        "--output-dir",
+        "${workspaceFolder}/target/robot_tests_result"
     ]
 }
--- a/tools/setup_robot_tests.sh
+++ b/tools/setup_robot_tests.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-# Copyright (c) 2025 Elektrobit Automotive GmbH
+# Copyright (c) 2026 Elektrobit Automotive GmbH
 #
 # This program and the accompanying materials are made available under the
 # terms of the Apache License, Version 2.0 which is available at


inf17101:
I will try this out and review this.

tools/setup_robot_tests.sh
```diff
@@ -0,0 +1,99 @@
+#!/bin/bash
+
+# Copyright (c) 2025 Elektrobit Automotive GmbH
```
inf17101:
```suggestion
# Copyright (c) 2026 Elektrobit Automotive GmbH
```

.vscode/settings.json
```diff
@@ -49,5 +49,12 @@
     "protobuf.diagnostics.namingConventions": false,
     "protobuf.includes": [
         "/workspaces/ankaios/ankaios_api/proto"
+    ],
+    "robotcode.robot.env": {
+        "ANK_BIN_DIR": "${workspaceFolder}/target/x86_64-unknown-linux-musl/debug"
+    },
+    "robotcode.robot.args": [
+        "--pythonpath",
+        "tests"
```
inf17101: We need to define the `output-dir` variable, otherwise a `results` folder is created in the project root on each test run.

```suggestion
    "robotcode.robot.args": [
        "--pythonpath",
        "tests",
        "--output-dir",
        "${workspaceFolder}/target/robot_tests_result"
```

tools/setup_robot_tests.sh
GabyUnalaq: My bad, I need to remember we are in a new year 😄

inf17101 (approved):
LGTM 👍

PR #710, opened 2026-03-04
Support multiple default config file paths in ank-cli

inf17101:
Issues: #707 

Changes to support `$HOME/.config/ankaios/ank.conf` and `/etc/ankaios/ank.conf` as default config paths for the ank-cli:

- introduces an array for supporting multiple default config file paths
- extends a swdd
- extends an utest for loading multiple default configs with first path taking precedence over the other
- updates the config file documentation to include now the two default paths for the ank-cli

# Definition of Done

### DIFF
--- a/common/src/config.rs
+++ b/common/src/config.rs
@@ -49,7 +49,26 @@ impl fmt::Display for ConversionErrors {
     }
 }
 
-pub fn handle_config<T: ConfigFile>(config_path: &Option<String>, default_path: &str) -> T {
+/// Handles the configuration file loading.
+///
+/// This function attempts to load the configuration from the provided `config_path`.
+/// If no path is provided, it will iterate over the `default_paths` and load the first
+/// existing configuration file. If no configuration file is found, it will return the
+/// default configuration.
+///
+/// # Type Parameters
+///
+/// * `T`: The type of the configuration file. Must implement `ConfigFile`.
+///
+/// # Arguments
+///
+/// * `config_path`: An optional path to the configuration file.
+/// * `default_paths`: A slice of default paths to search for the configuration file.
+///
+/// # Returns
+///
+/// The loaded configuration of type `T`.
+pub fn handle_config<T: ConfigFile>(config_path: &Option<String>, default_paths: &[&str]) -> T {
     match config_path {
         Some(config_path) => {
             let config_path = PathBuf::from(config_path);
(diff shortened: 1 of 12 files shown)


common/src/config.rs
```diff
@@ -49,7 +49,26 @@ impl fmt::Display for ConversionErrors {
...
+///
+/// # Arguments
+///
+/// * `config_path`: An optional path to the configuration file.
+/// * `default_paths`: A slice of default paths to search for the configuration file.
+///
+/// # Returns
+///
+/// The loaded configuration of type `T`.
+pub fn handle_config<T: ConfigFile>(config_path: &Option<String>, default_paths: &[&str]) -> T {
```
GabyUnalaq, line 62: I like having documentation in the code, but this is a bit too much, judging by how we documented the rest of the Ankaios code. I suggest trimming it down a bit:

```suggestion
/// Handles the configuration file loading.
///
/// This function attempts to load the configuration from the provided `config_path`.
/// If no path is provided, it will iterate over the `default_paths` and load the first
/// existing configuration file. If no configuration file is found, it will return the
/// default configuration.
///
/// ## Returns
///
/// The loaded configuration.
pub fn handle_config<T: ConfigFile>(config_path: &Option<String>, default_paths: &[&str]) -> T {
```

common/src/config.rs
```diff
@@ -60,20 +79,23 @@ pub fn handle_config<T: ConfigFile>(config_path: &Option<String>, default_path:
...
+                        "Loading config from default path '{}'",
+                        default_path.display()
+                    );
+                    return T::from_file(default_path)
+                        .unwrap_or_exit("Config file could not be parsed");
+                }
             }
+
+            log::debug!(
+                "No config file found at default paths '{:?}'. Continue with default config.",
```
GabyUnalaq, line 86: Because we now have more paths, I am afraid the message is a bit too large.
Up to you, but I suggest to remove the default paths values from the message, because they are present in the documentation and also in the `ank --help` output.

GabyUnalaq (commented):
Finished the review

common/src/config.rs
inf17101: Got your point. But I think there are only two paths in the CLI now and to be honest it's a debug print. We print much larger stuff like CompleteState redundantly in debug logs.

GabyUnalaq: Alright then

GabyUnalaq (approved):
LGTM 👍

PR #719, opened 2026-03-25
Prepare debian packages for Nexus 3 hosting

GabyUnalaq:
# Description

This PR adds script for creating and uploading Ankaios debian packages to Eclipse Nexus 3.

# Technicalities

1. the debian packages are created on Ubuntu 22.04, which makes them compatible with any distro that has at least glibc 2.35. This includes Ubuntu 22.04+, Debian 12+, RHEL 10+ and more.
2. The binary packages are uploaded to Nexus while the src package is published as an artifact for the moment.
3. The release triggers the build and publish of the debian packages automatically. Manual runs can also be made.

# Tasks

- [x] Implement scripts
- [x] Implement actions
- [x] Test the actions (Runs can be seen in my fork: [build&publish](https://github.com/GabyUnalaq/ankaios/actions/workflows/debian_build.yml) and [remove](https://github.com/GabyUnalaq/ankaios/actions/workflows/debian_remove_pkg.yml))
- [x] Add installation instructions to the docs

# Definition of Done


.github/scripts/debian/build_bin_pkgs.sh
```diff
@@ -0,0 +1,46 @@
+#!/bin/bash
+
+# Copyright (c) 2023 Elektrobit Automotive GmbH
```
windsource:
```suggestion
# Copyright (c) 2026 Elektrobit Automotive GmbH
```

.github/scripts/debian/build_src_pkg.sh
```diff
@@ -0,0 +1,267 @@
+#!/bin/bash
+
+# Copyright (c) 2023 Elektrobit Automotive GmbH
```
windsource:
```suggestion
# Copyright (c) 2026 Elektrobit Automotive GmbH
```

.github/scripts/debian/delete_pkg.py
```diff
@@ -0,0 +1,96 @@
+#!/usr/bin/env python3
+
+# Copyright (c) 2023 Elektrobit Automotive GmbH
```
windsource:
```suggestion
# Copyright (c) 2026 Elektrobit Automotive GmbH
```

.github/scripts/debian/publish_bin_pkgs.sh
```diff
@@ -0,0 +1,62 @@
+#!/bin/bash
+
+# Copyright (c) 2023 Elektrobit Automotive GmbH
```
windsource:
```suggestion
# Copyright (c) 2026 Elektrobit Automotive GmbH
```

windsource:
What about moving `.github/scripts/debian` to `tools/debian` such to only have one place for scripts? I know that the scripts in the debian folder are only used during CI builds but the same is valid for e.g. `tools/create_release.sh`.

tools/debian/build_src_pkg.sh
```diff
@@ -0,0 +1,267 @@
...
+Description: Eclipse Ankaios - full installation
+ Meta-package that installs all Ankaios components: server, agent, and CLI.
+EOF
+}
+
+write_service() {
+    local description="$1" bin_dir="$2" exec_name="$3" log_level="$4"
+    cat > "$BASE_DIR/debian/$exec_name.service" << EOF
+[Unit]
+Description=${description}
```
windsource, line 108:
```suggestion
Description=${description}
After=network-online.target
Wants=network-online.target
```

We should wait until the network is available.

tools/debian/build_src_pkg.sh
```diff
@@ -0,0 +1,267 @@
...
+
+write_copyright() {
+    cat > "$BASE_DIR/debian/copyright" <<EOF
+Format: https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/
+Upstream-Name: ankaios
+Upstream-Contact: ${MAINTAINER}
+Source: https://github.com/eclipse-ankaios/ankaios
+
+Files: *
+Copyright: 2023 Elektrobit Automotive GmbH
```
windsource:
```suggestion
Copyright: 2026 Elektrobit Automotive GmbH
```

tools/debian/build_src_pkg.sh
GabyUnalaq: Should this be implemented for all service files, including the ones set by the install script?

windsource: Yes, we should have that consistent.

doc/docs/usage/installation.md
```diff
@@ -89,7 +89,7 @@ curl -sfL https://github.com/eclipse-ankaios/ankaios/releases/download/<version>
 
 For available versions see the [list of releases](https://github.com/eclipse-ankaios/ankaios/tags).
 
-### Set the log level for `ank-server` and `ank-agent` services
+**Set the log level for `ank-server` and `ank-agent` services**
```
christoph-hamm: Just move this headline level for `####`.

doc/docs/usage/installation.md
```diff
@@ -111,7 +111,7 @@ To configure the log levels for `ank-server` and `ank-agent` during the installa
 
 Now, both services will output logs according to the specified log levels. If no explicit value was provided during installation, both services will default to `info` log level. You can always change the log level by updating the environment variables and reinstalling the services.
 
-### Uninstall Ankaios
+**Uninstall**
```
christoph-hamm: Same as above.

doc/docs/usage/installation.md
```diff
@@ -121,11 +121,56 @@ ank-uninstall.sh
...
+| `ank-agent` | Ankaios agent |
+| `ank` | Ankaios CLI |
+
+!!! note
+
+    The `ank-server` and `ank-agent` systemd services are started automatically after installation.
+
+The packages are compatible with Ubuntu 22.04+, Debian 12+ and other distributions based on glibc 2.35 or later.
+
+**Uninstall**
```
christoph-hamm: Also `####`.

.markdownlint.yml
```diff
@@ -17,5 +17,8 @@ MD033:
 # Disables code block style checks because of syntax for "!!! Note"
 MD046: false
 
+# Allow emphasis (bold) used as a visual sub-section label without TOC entry
+MD036: false
```
christoph-hamm: I would rather use a heading if you mean a heading.  I see, that you do not want this heading to be visible in the TOC. In my opionion there a two ways of doing this correctly:

* **Limit the TOC level globaly**: Add the `toc_depth: 3` option to the `mkdoc.yml` file. But this will also effect `unit_verification.md`.  Which could be fixed by improving the use of heading level there.
* **Limit the TOC level for a certain page**: You can add CSS to a markdown page, hidding TOC levels you do not want to be show: 

```
<style>
  .md-nav__title + ul nav nav {
    display: none;
  }
</style>
```

```suggestion
```

.gitignore
```diff
@@ -5,6 +5,7 @@ req_in
 dist
 logs
 vendor
+/debian
```
christoph-hamm, line 8: I do not think we need another folder for build ouput. We already have `build` and `target`. I guess we should just use `build/debian`

tools/install.sh
```diff
@@ -173,6 +175,8 @@ EOF
         $SVC_SUDO tee "$FILE_ANK_AGENT_SERVICE" >/dev/null << EOF
 [Unit]
 Description=Ankaios agent
+After=network-online.target
+Wants=network-online.target
```
christoph-hamm: We should use `network.target`. The `network-online.target` requires some routable IP. But ankaios-server and ankaios-agent will happily work if they can only use the loopback interface to communicate.  Also, it is recommendet to not use this target to liberally, as it can delay the boot of the system (see also in the systemd documentation [Network Configuration Synchronization Points](https://systemd.io/NETWORK_ONLINE/)).




```suggestion
After=network.target
Wants=network.target

[Service]
Environment="RUST_LOG=${INSTALL_ANK_SERVER_RUST_LOG}"
@@ -173,6 +175,8 @@ EOF
$SVC_SUDO tee "$FILE_ANK_AGENT_SERVICE" >/dev/null << EOF
[Unit]
Description=Ankaios agent
After=network.target
Wants=network.target
```

tools/generate_man_pages.sh
```diff
@@ -0,0 +1,135 @@
...
+.BR ank (1)
+INCLUDE_EOF
+    fi
+
+    help2man --no-info --section=1 --name="Ankaios CLI - ${cmd}" \
+        --include="$INCLUDE" \
+        "$WRAPPER" -o "$OUTPUT_DIR/man1/ank-${cmd}.1"
+    echo "Generated $OUTPUT_DIR/man1/ank-${cmd}.1"
+
+    rm -f "$WRAPPER" "$INCLUDE"
```
christoph-hamm: This is inconsistent. The INCLUDE you create above, is removed with `trap`, but the WRAPPER and the INCLUDE here are just normally removed. I would suggest to only use on INCLUDE file and create both, the INCLUDE and the WRAPPER file above and use trap to delete both. You do not need to fill the wrapper script immediately, you can still do this here. You could even move the creation of the temporary files to the top of the script.

tools/generate_man_pages.sh
```diff
@@ -0,0 +1,135 @@
...
+
+    # Wrapper script: help2man invokes the binary with --help / --version.
+    # Redirect --help to "ank <subcommand> --help".
+    # Patch --version to output "ank-<subcommand> <version>" so that help2man
+    # uses the correct program name in the .TH header (e.g. ANK-GET, not ANK).
+    cat > "$WRAPPER" << WRAPPER_EOF
+#!/bin/bash
+case "\$1" in
+  --help)    exec "${ANK}" ${cmd} --help ;;
+  --version) "${ANK}" --version | sed "s/^ank /ank-${cmd} /" ;;
```
christoph-hamm: Are you sure you need `exec` in the first case? You do not use it in the second case.

tools/debian/build_src_pkg.sh
```diff
@@ -0,0 +1,269 @@
...
+EOF
+}
+
+write_service() {
+    local description="$1" bin_dir="$2" exec_name="$3" log_level="$4"
+    cat > "$BASE_DIR/debian/$exec_name.service" << EOF
+[Unit]
+Description=${description}
+After=network-online.target
+Wants=network-online.target
```
christoph-hamm: As mentioned for the install script, this should be `network.target`.
```suggestion
After=network.target
Wants=network.target
```

tools/debian/build_src_pkg.sh
```diff
@@ -0,0 +1,269 @@
...
+RELEASE_URL="https://github.com/eclipse-ankaios/ankaios/releases/tag/v$ANKAIOS_VERSION"
+
+write_control() {
+    cat > "$BASE_DIR/debian/control" <<'EOF'
+Source: ankaios
+Section: embedded
+Priority: optional
+Maintainer: Eclipse Ankaios <ankaios-dev@eclipse.org>
+Build-Depends: debhelper-compat (= 13),
+ rustc-1.89 (>= 1.89.0) | rustc (>= 1.89.0),
```
christoph-hamm: I do not think you need the version check on the left side.

tools/debian/build_src_pkg.sh
```diff
@@ -0,0 +1,269 @@
...
+EOF
+}
+
+write_rules() {
+    cat > "$BASE_DIR/debian/rules" <<'EOF'
+#!/usr/bin/make -f
+
+# Use Rust 1.89 versioned packages from apt.
+export RUSTC = /usr/bin/rustc-1.89
+export CARGO = /usr/bin/cargo-1.89
```
christoph-hamm, line 141: This does not fit the make dependencies above. They either want `rustc-1.89` or `rustc(>= 1.89.0)`. If we also support the usage of any rustc >= 1.89.0, we should use `/usr/bin/rustc`. Same for cargo below.
```suggestion
export RUSTC = /usr/bin/rustc
export CARGO = /usr/bin/cargo
```

tools/debian/build_src_pkg.sh
```diff
@@ -0,0 +1,269 @@
...
+export CARGO_HOME = $(CURDIR)/debian/.cargo
+
+# Derive the Rust target triple from DEB_HOST_GNU_TYPE.
+DEB_HOST_GNU_TYPE := $(shell dpkg-architecture -qDEB_HOST_GNU_TYPE)
+RUST_HOST := $(subst -linux-gnu,-unknown-linux-gnu,$(DEB_HOST_GNU_TYPE))
```
christoph-hamm: Why not just ask `rustc` for the host tuple?
```suggestion
RUST_HOST := $(rustc --print host-tuple)
```

tools/debian/build_src_pkg.sh
```diff
@@ -0,0 +1,269 @@
...
+Maintainer: Eclipse Ankaios <ankaios-dev@eclipse.org>
+Build-Depends: debhelper-compat (= 13),
+ rustc-1.89 (>= 1.89.0) | rustc (>= 1.89.0),
+ cargo-1.89 | cargo (>= 1.89.0),
+ protobuf-compiler,
+ help2man
+Standards-Version: 4.6.2
+Homepage: https://github.com/eclipse-ankaios/ankaios
+
+Package: ank-server
```
christoph-hamm, line 59: I think it would be more consistent to call the packages `ankaios-server`, `ankaios-agent` and `ankaios-cli`. E.g. the package for the helix editor is also `helix`, also the command is `hx`.

.github/workflows/build.yml
```diff
@@ -159,8 +159,8 @@ jobs:
           name: code-coverage
           path: target/llvm-cov/html
 
-  build_linux_amd64_debian_package:
-    name: Build Linux amd64 debian package
+  build_linux_amd64:
```
christoph-hamm, line 162: The key does not fit the name below.

tools/debian/delete_pkg.py
```diff
@@ -0,0 +1,96 @@
...
+    for item in items:
+        component_id = item["id"]
+        name = item.get("name", "?")
+        ver = item.get("version", "?")
+        arch = item.get("group", "?")  # Nexus uses "group" for architecture in apt repos
+
+        if architecture != "*" and arch != architecture:
+            continue
+
+        print(f"Deleting {name} {ver} {arch} ({component_id})...")
```
christoph-hamm, line 87: I am not sure about this message. If `name`, `var` and `arch` are all unkown, this would get not readable. But can they even be unkown?

tools/debian/delete_pkg.py
```diff
@@ -0,0 +1,96 @@
...
+        del_req = urllib.request.Request(
+            f"{NEXUS_URL}/service/rest/v1/components/{component_id}",
+            method="DELETE",
+            headers=headers,
+        )
+        with urllib.request.urlopen(del_req) as resp:
```
christoph-hamm: We completely ignore the return code. I think we should at least log an error, if something went wrong. I think we should also return an error at the end, if one package could not be deleted. I would still continue to try deleting the remaining packages.

tools/debian/delete_pkg.py
```diff
@@ -0,0 +1,96 @@
...
+
+username = os.environ["REPO_TOKEN_USERNAME"]
+password = os.environ["REPO_TOKEN_PASSWORD"]
+package = os.environ.get("PACKAGE", "*")
+version = os.environ.get("VERSION", "*")
+architecture = os.environ.get("ARCHITECTURE", "*")
+revision = os.environ.get("REVISION", "")
+repository = os.environ.get("REPOSITORY", "ankaios-apt")
+
+full_version = f"{version}-{revision}" if revision else version
```
christoph-hamm, line 48: What exactelly happens, if `revsion` is set, but `version` isn't?  I doubt that `*-2`, e.g., is an valid parameter which matches anything.

.github/workflows/debian_remove_pkg.yml
```diff
@@ -0,0 +1,44 @@
...
+        required: false
+        type: string
+        default: "*"
+      architecture:
+        description: "Architecture (amd64, arm64). Use * for all."
+        required: false
+        type: string
+        default: "*"
+      revision:
+        description: "Build revision (e.g. 1). Leave empty to ignore."
```
christoph-hamm: Is this correct? If I give version `0.9.2` and no revision, would this also delete version `0.9.2-2`?

.github/workflows/release.yml
```diff
@@ -103,3 +103,10 @@ jobs:
       - name: Upload quality manifest
         run: |
           gh release upload ${{ github.ref_name }} ${{ steps.collect_quality_artifacts.outputs.manifest_file }}
+  debian-packages:
+    name: Build and publish Debian packages
+    needs: release
+    permissions:
+      contents: read
+    uses: ./.github/workflows/debian_build.yml
+    secrets: inherit
```
christoph-hamm, line 112: Will this set the input `ankaios_version` correctly?

.github/workflows/debian_build.yml
```diff
@@ -0,0 +1,108 @@
...
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+
+      - name: Resolve version
+        id: vars
+        run: |
+          version="${{ inputs.ankaios_version }}"
+          if [ -z "$version" ]; then
```
christoph-hamm, line 32: Does this make sense? The parameter `version` is marked as required above.

.github/workflows/debian_build.yml
```diff
@@ -0,0 +1,108 @@
...
+      - name: Install build dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y \
+            debhelper \
+            dpkg-dev \
+            rustc-1.89 \
+            cargo-1.89 \
+            protobuf-compiler \
+            help2man
```
christoph-hamm: Wouldn't it be better to install the build dependencies as stored in the dsc file, using `apt-get build-dep`?  Now we have two places you need to change, once you at a build dependency, the `build_src_pkg.sh` to include the in the src package and here to actually install them.

christoph-hamm (changes requested):

.gitignore
GabyUnalaq: The debian source package generation requires a debian folder inside of the project that is to be turned into the package, thus we need a debian folder in the root. That is why I added it here.

tools/debian/build_src_pkg.sh
GabyUnalaq: After a discussion, we decided to leave the pinpointed 1.89 version, and this included fixing the dependencies list to only install that.

tools/debian/build_src_pkg.sh
GabyUnalaq: After a discussion, we decided to leave the pinpointed 1.89 version. This will fix the next finding as well.

.github/workflows/build.yml
GabyUnalaq: What do you mean?

tools/debian/delete_pkg.py
GabyUnalaq: They can never be unknown.

tools/debian/delete_pkg.py
GabyUnalaq: Added handling

tools/debian/delete_pkg.py
GabyUnalaq: Added check if the revision is set without version.

.github/workflows/debian_remove_pkg.yml
GabyUnalaq: It will not, the version check is a direct match.

.github/workflows/release.yml
GabyUnalaq: `debian_build.yml` does not have any inputs, the version is not set directly, rather in the workflow itself it resolves it directly from `GITHUB_REF_NAME`.

.github/workflows/debian_build.yml
GabyUnalaq: It is required only for the workflow_dispatch and it does not apply to the workflow_call option.

.github/workflows/debian_build.yml
GabyUnalaq: Sounds good! There are some problems:
1. I've tried the `apt-get build-dep` in a container with the dsc file, and it didn't work as expected (`E: You must put some 'deb-src' URIs in your sources.list`)
2. Apparently there is a tool which does this directly: `mk-build-deps` which is part of `devscripts`. While testing it, it is a lot of dependencies for a simple enough task
3. Simple enough task means I can do it manually:
    `deps=$(grep '^Build-Depends:' <>.dsc | sed 's/^Build-Depends: //')`
    and this is what I'll do. I've tested it and it works as expected.

.github/workflows/debian_remove_pkg.yml
christoph-hamm: But then the description is misleading.

.github/workflows/debian_build.yml
christoph-hamm: OK

.github/workflows/debian_build.yml
christoph-hamm: You can also only use `sed`, with `cat <>.dsc | sed -n 's/^Build-Depends: //p'` or  `sed -n 's/^Build-Depends: //p' <>.dsc`

.github/workflows/build.yml
christoph-hamm: I have not idea either what I meant. Everything looks good here. Forget about it.

tools/debian/delete_pkg.py
christoph-hamm: But in line 75-77 there is special handling setting them to "?" if they are not set.

GabyUnalaq: The normal Nexus message, from what I've seen, always contains those fields, but just in case, the code will not fail if something goes wrong or that is not always the case.

.github/workflows/debian_build.yml
GabyUnalaq: Changed it to the second proposed option.

.github/workflows/debian_remove_pkg.yml
GabyUnalaq: Fixed the description

christoph-hamm (approved):

tools/debian/build_src_pkg.sh
windsource: > I think it would be more consistent to call the packages `ankaios-server`, `ankaios-agent` and `ankaios-cli`. E.g. the package for the helix editor is also `helix`, also the command is `hx`.

Actually the helix command on Arch Linux is `helix` as `hx` has been reserved already. So from my point of having the meta package being called `ankaios` and the others `ank`, `ank-server` and `ank-agent` would be fine as the executables are called like that as well and it is not too long.

windsource (approved):
LGTM 👍

PR #730, opened 2026-04-27
Support building of Arch Linux packages

christoph-hamm:
Issues: #

# Definition of Done


tools/aur/publish_to_aur.sh
```diff
@@ -0,0 +1,35 @@
...
+ASSETS=("ank-server.service" "ank-agent.service" "ankaios-cli.install")
+
+if [ -z "${ANKAIOS_VERSION}" ]; then
+    echo "Error: ANKAIOS_VERSION environment variable is not set."
+    exit 1
+fi
+
+tmp_dir="$(mktemp -d)"
+echo $tmp_dir
+trap 'rm -rf "$tmp_folder"' EXIT
```
krucod3:
```suggestion
trap 'rm -rf "$tmp_dir"' EXIT
```

tools/aur/publish_to_aur.sh
```diff
@@ -0,0 +1,35 @@
...
+
+cd "$SCRIPT_DIR"
+for package in ankaios ankaios-bin ankaios-git; do
+    git clone "$AUR_REPO_BASE/$package.git" "$tmp_dir/$package"
+    git checkout -b master || true # AUR only accepts the master branch
+
+    cd "$tmp_dir/$package"
```
krucod3: Isn't the order wrong here?

tools/aur/ankaios-cli.install
krucod3, line 1: Indentation is not unified (lines 3-4 with tabs).

Missing new line at the end.

tools/aur/publish_to_aur.sh
```diff
@@ -0,0 +1,35 @@
...
+AUR_REPO_BASE="${AUR_REPO_BASE:-ssh://aur@aur.archlinux.org}"
+ASSETS=("ank-server.service" "ank-agent.service" "ankaios-cli.install")
+
+if [ -z "${ANKAIOS_VERSION}" ]; then
+    echo "Error: ANKAIOS_VERSION environment variable is not set."
+    exit 1
+fi
+
+tmp_dir="$(mktemp -d)"
+echo $tmp_dir
```
krucod3:
```suggestion
echo "$tmp_dir"
```

tools/aur/PKGBUILD-ankaios-bin.m4
```diff
@@ -0,0 +1,66 @@
...
+    conflicts=(ankaios-cli)
+
+    install=ankaios-cli.install
+    install -Dm755 -t "$pkgdir"/usr/bin/ ank
+    install -Dm644 -t "$pkgdir"/etc/ankaios ank.conf
+    install -Dm644 -t "$pkgdir"/usr/share/man/man1 man1/*
+}
+
+package_ankaios-bin() {
+    pkgdesc="Meta-package to installs all components of Eclipse Ankaios"
```
krucod3:
```suggestion
    pkgdesc="Meta-package to install all components of Eclipse Ankaios"
```

tools/aur/PKGBUILD-ankaios.m4
```diff
@@ -0,0 +1,69 @@
...
+package_ankaios-cli() {
+    pkgdesc="A command line tool for communicating with the API of the Eclipse Ankaios server"
+    install=ankaios-cli.install
+    install -Dm755 -t "$pkgdir"/usr/bin/ "ankaios-$pkgver/target/$(rustc --print host-tuple)/release/ank"
+    install -Dm644 -t "$pkgdir"/etc/ankaios/ "ankaios-$pkgver"/ank/config/ank.conf
+    install -Dm644 -t "$pkgdir"/usr/share/man/man1 "ankaios-$pkgver"/build/man/man1/*
+}
+
+package_ankaios() {
+    pkgdesc="Meta-package to installs all components of Eclipse Ankaios"
```
krucod3:
```suggestion
    pkgdesc="Meta-package to install all components of Eclipse Ankaios"
```

tools/aur/PKGBUILD-ankaios.m4
```diff
@@ -0,0 +1,69 @@
...
+    export RUSTUP_TOOLCHAIN=stable
+    export CARGO_TARGET_DIR=target
+    export CFLAGS+=" -ffat-lto-objects"
+    cargo build --frozen --release --target host-tuple
+    ./tools/generate_man_pages.sh "target/$(rustc --print host-tuple)/release/" build/man/
+}
+
+package_ankaios-server() {
+    pkgdesc="The server application of Eclipse Ankaios"
+
```
krucod3, line 39: Provides and conflicts are missing here.

tools/aur/ank-agent.service
```diff
@@ -0,0 +1,9 @@
+[Unit]
+Description=Ankaios agent
```
krucod3, line 2: Given the issues that we observed lately, shouldn't add some dependencies here too:

```suggestion
Description=Ankaios agent
After=network.target ank-server.service
```

Edit: we shouldn't put a `Wants=ank-server.service` inside as it could be an agent only node.

tools/aur/ank-server.service
```diff
@@ -0,0 +1,9 @@
+[Unit]
+Description=Ankaios server
```
krucod3, line 2:
```suggestion
Description=Ankaios server
After=network.target 
```

krucod3 (changes requested):
Initial review done, findings documented.

tools/aur/PKGBUILD-ankaios.m4
christoph-hamm: TLDR: This is how it should be.

This is basically the PKGBUILD as you would find it in an official package. It build from the official released source, hence it has not suffix like `-bin` or `-git`. Hence these package just provide what is their name and also are not in conflict with anything else. 

The `-bin` and `-git` package are alternatives. 
They have to state for which package you could use them (the `provides` part) and also with which package they conflict.

tools/aur/publish_to_aur.sh
christoph-hamm: Yes. Fixed.

tools/aur/ankaios-cli.install
christoph-hamm: Fixed

tools/aur/PKGBUILD-ankaios.m4
krucod3: point taken.

tools/aur/README.md
```diff
@@ -0,0 +1,39 @@
...
+| ------ | --------- |
+| `publish_to_aur.sh` | Clones AUR repositories, generates `PKGBUILD` from templates, updates checksums and `.SRCINFO`, commits, and pushes changes |
+| `PKGBUILD-ankaios.m4` | Template for the release-from-source AUR package |
+| `PKGBUILD-ankaios-bin.m4` | Template for the prebuilt-binary AUR package |
+| `PKGBUILD-ankaios-git.m4` | Template for the git-head AUR package |
+| `ank-server.service`, `ank-agent.service`, `ankaios-cli.install` | Asset files copied into each AUR package repository |
+
+## Required environment
+
+- `ANKAIOS_VERSION` must be set (for example `v1.0.0`)
```
krucod3: This is later prefixed with a v:

v$pkgver


this will result in `vv1.0.1`

tools/aur/ank-agent.service
krucod3, line 1: These service files are now triplicated 🤣 . We have them in the install script, in the Debian package and here.
Soon we will also have them in the meta-ankaios.

Was there a reason to not put them in a shared folder under `tools`?

tools/aur/PKGBUILD-ankaios-bin.m4
```diff
@@ -0,0 +1,66 @@
...
+    install -Dm644 -t "$pkgdir"/usr/share/man/man8 man8/ank-server.8
+}
+
+package_ankaios-agent-bin() {
+    pkgdesc="An agent running on each node of an Eclipse Ankaios cluster"
+    provides=(ankaios-agent)
+    conflicts=(ankaios-agent)
+    optdepends=(
+      'podman: for running podman workloads'
+      'nerdctl: for running containerd workloads'
```
krucod3, line 46: We would also need containerd itself.

tools/aur/README.md
christoph-hamm: fixed

tools/aur/ank-agent.service
christoph-hamm: I am not sure about this. On one hand these are files which depend on the Linux distribution, and hence I would put them separate, on the other hand at the moment they are all the same.

windsource: I also rather think, that the service files are closely related to the specific installation method. For example the install script installs to `/usr/local/bin/` but debian packages not.

tools/aur/PKGBUILD-ankaios-bin.m4
christoph-hamm: I would not include it. The [PKGBUILD documentations](https://wiki.archlinux.org/title/PKGBUILD#depends) states: "The depends array should list all direct first level dependencies even when some are already declared transitively".

Containerd might only be declared "transitively", but I think it is not first level. To my understanding we do not directly interact with containerd.

tools/aur/ank-agent.service
krucod3: So one out of the four requires a different path. In all three other installation methods that are under our control we use the same destination `/usr/bin/`. 

In my eyes this is code duplication that would easily result in diverging configurations. We see this already now, when we added the `After=network.target` which was in the Debian package, but not here and the `After=network.target ank-server.service` in the agent service which is not here, but on in the Debian service.

tools/aur/PKGBUILD-ankaios-bin.m4
krucod3: Alright, containerd is a hard dependency of nerdctl so it should be fine (see https://archlinux.org/packages/extra/x86_64/nerdctl/).

tools/aur/ank-agent.service
christoph-hamm: But conventions how you handle such things can be easily different between different distributions.

I see it like this: normally the service files are not provided by the developer of the software, but added by the package maintainer of the distribution. Hence it is normal for service files to be divert.

I am also not sure if we need to fix this now. These files are only some sideshow and should not delay our upcoming release.

krucod3: Now all duplicated files have the same content. Let's extract to a common place the next time they diverge ...

krucod3 (approved):
LGTM 👍

tools/aur/ank-agent.service
christoph-hamm: OK.

PR #736, opened 2026-05-12
Add Sonarqube GitHub action

krucod3:
Issues: #

# Definition of Done

### DIFF
--- a/.github/workflows/sonarcloud.yml
+++ b/.github/workflows/sonarcloud.yml
@@ -100,6 +100,10 @@ jobs:
           lcov_to_sonar(sys.argv[1], sys.argv[2])
           PYEOF
 
+      - name: Install scanner dependencies
+        # this should be added to the base image at some point, but for now we need to install it in the workflow
+        run: apt-get update && apt-get install -y --no-install-recommends unzip
+
       - name: SonarCloud Scan
         uses: SonarSource/sonarqube-scan-action@59db25f34e16620e48ab4bb9e4a5dce155cb5432 # v8.0.0
         env:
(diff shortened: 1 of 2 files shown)


.github/workflows/sonarcloud.yml
```diff
@@ -0,0 +1,102 @@
...
+              file_lines = []
+
+              with open(lcov_path) as f:
+                  for raw in f:
+                      line = raw.strip()
+                      if line.startswith('SF:'):
+                          if current_file and file_lines:
+                              lines_out.append(f'  <file path="{html.escape(current_file)}">')
+                              lines_out.extend(file_lines)
+                              lines_out.append('  </file>')
```
christoph-hamm, line 67: These lines are duplicate from the lines below. I also think, that we do not need them, as the the `end_of_record` line is mandatory (see the documentation of the format https://github.com/linux-test-project/lcov/blob/07a1127c2b4390abf4a516e9763fb28a956a9ce4/man/geninfo.1#L989)

christoph-hamm (changes requested):

.github/workflows/sonarcloud.yml
krucod3: Here is the Copilot Answer: 

But they are not redundant — they serve different purposes in the LCOV parsing logic:

Lines 65-67 (inside if line.startswith('SF:')) flush the previous file's data when a new SF: record starts — handling the case where there's no explicit end_of_record between files.
Lines 86-88 (inside elif line == 'end_of_record') flush the current file's data when the record is explicitly terminated.

christoph-hamm: I want to argue about this. I do not care about some Github action we will probably never touch again.

christoph-hamm (approved):

Issue #740, opened 2026-05-15
Mutating Admission Hooks


krucod3:
## Description

Add mutating admission hooks to Ankaios.

The target is to use the admission hooks feature to implement plugins for various use-cases for Ankaios:

* validate workload configuration
* handle automatically security hardening configurations
* automatically add resource usage limitations to a runtimeConfig
* schedule a workload automatically to a free node if the workload is tagged accordingly
* setup networking before a workload is scheduled

## Goals

* Enable flexible plugins executed in Ankaios before a workload is started.
* Add state update rejection possibilities before the update is applied.

### Summary

**To be filled when the final solution is sketched.**

### Tasks

- [ ] Task 1
- [ ] Task 2
- [ ] ...

krucod3:
We still need to clarify some topics:

* Hook configuration possibilities: 
I think it would be best to provide at some point the possibility of quick filtering of applicable workloads, s.t. not each hook is called for each workload. As a start we can just do exactly this and regard it as a default behavior, but we could later add filtering on workload name (e.g. regex to match a prefix, suffix, etc.) or distinct tags.

Assumptions:
* Security: the current assumption is that the filesystem is initially secured in some way (e.g. dm-verity, or dm-integrity and a secure book) and validating the executables on each run is not necessary.
* Performance: the current assumption is that staring an executable for each hook call is acceptable. If we want to have daemon hooks an additional security layer may be needed depending on the communication type and the protocol must be more complex. Assuming a relatively infrequent calls to the hooks and chaining up to 10-20 hooks at most calling a new process for each hook call is acceptable. Still, this could be later changed with an additional attribute in the hook config that introduces a new hook type.
* no need to differentiate between admission and mutating hooks. Both use cases are supported with one API and it is up to the system designer to decide how to use them.

krucod3:
Done.

PR #741, opened 2026-05-15
Mutating Admission Hooks 

krucod3:
Issue: #740 

This PR adds mutating admission hooks to Ankaios.

# Definition of Done


server/doc/swdesign/README.md
```diff
@@ -49,6 +49,30 @@ Considered alternatives:
...
+#### Admission hook response optimization via byte comparison
+`swdd~server-admission-hook-skip-unchanged-response~1`
+
+Status: approved
+
+When an admission hook returns its response, the server compares the raw protobuf output bytes against the input bytes. If they are identical, the server skips decoding and conversion entirely.
+
+Rationale:
+
+Protobuf encoding is deterministic for the same message, so identical bytes guarantees an identical message. A byte-level comparison (`memcmp`) is extremely cheap and avoids unnecessary protobuf decoding, object allocation, and type conversion when a hook approves the workloads without modifying them. This is the expected common case, as most hooks will only validate or reject, not mutate.
```
christoph-hamm: To my knowledge protobuf encoding is not deterministic.

agent/src/runtime_manager.rs
```diff
@@ -2411,19 +2361,6 @@ mod tests {
...
-        //         response_content: Some(ResponseContent::CompleteStateResponse(Box::new(
-        //             generate_test_complete_state(vec![generate_test_workload_spec_with_param(
-        //                 AGENT_NAME.to_string(),
-        //                 WORKLOAD_1_NAME.to_string(),
-        //                 RUNTIME_NAME.to_string(),
-        //             )])
-        //             .into(),
-        //         ))),
-        //     })
-        //     .await;
```
inf17101, line 2420: There is comment out test code.

inf17101 (commented):
I reviewed the code and hooks implementation already to get familiar with the hooks implementation for my next task. I know the PR is not ready for review, but found one simple utest comment out code, that needs to be fixed.

I also noticed that the process is spawned by the user executing Ankaios. Some hooks might require root privileges. This implies that Ankaios must run as root or we introduce a user context. In my opinion, the this implementation is really good and sufficient, because a hook requiring root privileges can also run as a high privileged daemon and a hook sends a request to this daemon if reconfiguration is required. This contributes to privilege separation and enables also the service as a reusable core service part of the platform instead of a direct hook.

server/src/ankaios_server/server_state.rs
```diff
@@ -165,7 +186,27 @@ fn include_both_state_and_substate_filters(filters: &mut Vec<String>) {
...
+    /// The effective state reflects workloads after config rendering and admission hook mutations.
+    fn build_effective_state(&self) -> StateSpec {
+        let workloads = self
+            .rendered_workloads
+            .iter()
+            .map(|(name, wl_named)| (name.clone(), wl_named.workload.clone()))
+            .collect();
+        StateSpec {
+            api_version: self.state.desired_state.api_version.clone(),
+            workloads: WorkloadMapSpec { workloads },
```
inf17101, line 203: I noticed that the order of the workloads in the effective state is different compared to the desiredState. I think the order is lost when generating the rendered workloads, so we can leave it as it is for the implementation or sort by the workload name again when creating the effective state.

Short example output:

```yaml
desiredState:
  apiVersion: v1
  workloads:
    workload_update_1:
      agent: agent_A
      restartPolicy: NEVER
      tags:
        project: team
      runtime: podman
      runtimeConfig: |-
        image: docker.io/alpine:latest
        commandArgs: ["sh", "-c", "echo 'workload_update_1 running'; sleep 3800"]
    workload_update_2:
      agent: agent_A
      restartPolicy: NEVER
      tags:
        project: team
      runtime: podman
      runtimeConfig: |-
        image: docker.io/alpine:latest
        commandArgs: ["sh", "-c", "echo 'workload_update_2 running'; sleep 3600"]
workloadStates:
  agent_A:
    workload_update_2:
      15076c14041dbe4a313c9d70793f45ded6b6137259e4a23bf46db89235cb7f77:
        additionalInfo: ''
        state: Running
        subState: Ok
    workload_update_1:
      0331c9662aa26653ac4418b4a3bfd22e04e6a211a57c3bc920f126cb3415d68e:
        additionalInfo: ''
        state: Running
        subState: Ok
agents:
  agent_A:
    status:
      cpuUsage: 1
      freeMemory: 653254656
effectiveState:
  apiVersion: v1
  workloads:
    workload_update_2:
      agent: agent_A
      restartPolicy: NEVER
      tags:
        project: team
      runtime: podman
      runtimeConfig: |-
        image: docker.io/alpine:latest
        commandArgs: ["sh", "-c", "echo 'workload_update_2 running'; sleep 3600"]
    workload_update_1:
      agent: agent_A
      restartPolicy: NEVER
      tags:
        project: team
      runtime: podman
      runtimeConfig: |-
        image: docker.io/alpine:latest
        commandArgs: ["sh", "-c", "echo 'workload_update_1 running'; sleep 3800"]
```

server/src/ankaios_server.rs
```diff
@@ -513,6 +525,26 @@ impl AnkaiosServer {
...
+                                    self.to_agents
+                                        .error(
+                                            request_id,
+                                            format!(
+                                                "Update rejected by admission hook '{}': '{}'",
+                                                hook, reason
+                                            ),
+                                        )
+                                        .await
+                                        .unwrap_or_illegal_state();
```
christoph-hamm: Should we add a test for this?

server/src/ankaios_server/admission_hook_registry.rs
```diff
@@ -0,0 +1,556 @@
...
+use crate::ankaios_server::rendered_workloads::RenderedWorkloads;
+use crate::ankaios_server::server_state::{AddedDeletedWorkloads, UpdateStateError};
+use crate::server_config::AdmissionHook;
+
+#[cfg_attr(test, mockall::automock)]
+pub trait AdmissionHookMutator {
+    fn mutate_with_admission_hooks(
+        &self,
+        added_deleted_workloads: &mut AddedDeletedWorkloads,
+    ) -> Result<(), UpdateStateError>;
```
christoph-hamm: Is this really a good interface? Why not:
```suggestion
    fn mutate_with_admission_hooks(
        &self,
        added_deleted_workloads: AddedDeletedWorkloads,
    ) -> Result<AddedDeletedWorkloads, UpdateStateError>;
```

Also has the benefit that in the error case the caller has no reference to the `added_deleted_workloads`, where the signature of this functions give me no clue if it still contains any meaningful data.

ankaios_api/proto/ank_base.proto
```diff
@@ -181,6 +181,7 @@ message CompleteState {
     State desiredState = 1; // The state the user wants to reach.
     WorkloadStatesMap workloadStates = 2; // The current execution states of the workloads.
     AgentMap agents = 3; // The agents currently connected to the Ankaios cluster.
+    State effectiveState = 4; // The rendered state with expanded templates after admission hooks mutations. Reflects the actual workload configurations sent to agents.
```
christoph-hamm: Do we realy need a `State` here? Only the workloads are altered through config expanding and admission hooks. A `WorkloadMap` should be sufficient.

server/src/ankaios_server/server_state.rs
```diff
@@ -165,7 +186,27 @@ fn include_both_state_and_substate_filters(filters: &mut Vec<String>) {
...
+            // Don't fill the configs as they are not touched.
+            // Nevertheless use the state to allow adding other fields in the future
```
christoph-hamm: This explanation should not be here in the source code. This should be either in the proto file where the effective state is defined or in a design decision.

server/src/ankaios_server.rs
christoph-hamm, line 805: This should handle the effective state as well, so you can also subscribe to it via events.

server/src/ankaios_server/admission_hook_registry.rs
```diff
@@ -0,0 +1,556 @@
...
+
+                *added_deleted_workloads =
+                    AddedDeletedWorkloads::try_from(mutated_proto).map_err(|err| {
+                        UpdateStateError::ResultInvalid(format!(
+                            "Failed to convert output from hook '{}': {err}",
+                            hook.name
+                        ))
+                    })?;
+            }
+        }
```
christoph-hamm: I think it would be better to forward stderr to logs. This allow hook developers to also print information in the success case. Maybe giving explanation why certain things have been dropped/added/alterd.

agent/src/runtime_manager.rs
krucod3: Yes, I found some commented out test code checked if we need it and deleted it with this PR.

server/src/ankaios_server/server_state.rs
krucod3: The desired state is also not ordered and is also not maintaining the order of the input manifest.
We also do not have an ordering in the workload states.

I would leave cosmetic changes out until a use-case requires them.

server/doc/swdesign/README.md
krucod3: You are completely right.

Updated.

server/src/ankaios_server/server_state.rs
krucod3: Done.

ankaios_api/proto/ank_base.proto
krucod3: Extended comment.

server/src/ankaios_server/admission_hook_registry.rs
krucod3: In my head (🤣), the current interface means, small or no changes. When I see an interface as the proposed one, I think about a completely reworked object that has little to do with the initial one. 

Probably only a thing in my head, but I would like to keep it as as. Let me know if you nevertheless insists on the change to continue the discussion.

server/src/ankaios_server/admission_hook_registry.rs
krucod3: Good point. Update done.

server/src/ankaios_server.rs
krucod3: I collapsed it one branch now. In my eyes, the test is not bringing a lot of benefit and only increasing the complexity. With the new implementation, the thing that can go wrong and is not well tested is the tracing ...

server/src/ankaios_server.rs
krucod3: Done.

server/src/ankaios_server/server_state.rs
inf17101: Ok yes let's keep it.

server/src/ankaios_server.rs
christoph-hamm: OK

server/src/ankaios_server/admission_hook_registry.rs
christoph-hamm: OK. Your probable right, especially as the name and types imply the change.

ankaios_api/proto/ank_base.proto
christoph-hamm: OK

server/src/ankaios_server/hooks_registry.rs
```diff
@@ -0,0 +1,555 @@
...
+
+            let executable = hook.path.join(&hook.name);
+            let mut child = Command::new(&executable)
+                .stdin(Stdio::piped())
+                .stdout(Stdio::piped())
+                .stderr(Stdio::piped())
+                .spawn()
+                .map_err(|err| UpdateStateError::MutatingHookError {
+                    hook: hook.name.clone(),
+                    reason: format!("Failed to execute hook '{}': {err}", hook.name),
```
christoph-hamm: We could make the live of people easier by printing the whole path here.
```suggestion
                    reason: format!("Failed to execute hook '{}': {err}", executable.display()),
```

server/src/ankaios_server/hooks_registry.rs
```diff
@@ -0,0 +1,555 @@
...
+                        "Hook '{}' rejected the change: {}",
+                        hook.name,
+                        stderr.trim()
+                    ),
+                });
+            }
+
+            // [impl->swdd~server-traces-mutating-hook-stderr~1]
+            log::debug!(
+                "Mutating hook '{}' completed successfully: {}",
```
christoph-hamm: Is it OK to have lines without level and timestamp? This happens if the hook logs multiple lines. The logger handles this with 4 spaces at the beginning of lines, but are we using this somewhere else already, or do we not want this.

Also, I think this line looks strange if the hook is not logging anything.

christoph-hamm (approved):
I have only the two minor findings in the comments above. Once they are fixed or rejected this PR can be merged.

server/src/ankaios_server/hooks_registry.rs
christoph-hamm: Line has been changed to contain "Output" before the colon. Clearly indicates it is followed by logs of the hook.

Also clarified multi line logs already used.

krucod3:
All findings fixed.

PR #749, opened 2026-05-19
Fix debian publishing condition and src package upload

GabyUnalaq:
# Description

This PR fixes 2 issues:
- The debian package gets built and published on all releases. It should be only on big releases, not on "pre" and release candidates;
- The debian source package is prepared and uploaded as an artifact, but it wasn't uploaded as a release artifact as well.

# Definition of Done

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -112,7 +112,26 @@ jobs:
   debian-packages:
     name: Build and publish Debian packages
     needs: release
+    if: ${{ !contains(github.ref_name, '-') }}
     permissions:
       contents: read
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -112,7 +112,26 @@ jobs:
   debian-packages:
     name: Build and publish Debian packages
     needs: release
+    if: ${{ !contains(github.ref_name, '-') }}
```
krucod3, line 115: So everything with a dash. Should be fine, it will catch even some other things 👍

krucod3 (approved):
LGTM 👍

PR #759, opened 2026-06-26
Cleanup runtime interface

christoph-hamm:
Remove generics from runtime interface.
This should show that it is possible to move runtimes to shard objects plugins.

# Definition of Done

### DIFF
--- a/agent/src/runtime_connectors/podman_kube/podman_kube_runtime.rs
+++ b/agent/src/runtime_connectors/podman_kube/podman_kube_runtime.rs
@@ -414,7 +414,7 @@ impl RuntimeConnector for PodmanKubeRuntime {
 
 #[async_trait]
 // [impl->swdd~podman-kube-implements-runtime-state-getter~1]
-impl RuntimeStateGetter<RuntimeWorkloadId> for PodmanKubeRuntime {
+impl RuntimeStateGetter for PodmanKubeRuntime {
     async fn get_state(&self, id: &RuntimeWorkloadId) -> ExecutionStateSpec {
         let instance_name = match Self::instance_name_from_runtime_id(id) {
             Ok(instance_name) => instance_name,
--- a/agent/src/runtime_connectors/podman_kube/podman_kube_runtime.rs
+++ b/agent/src/runtime_connectors/podman_kube/podman_kube_runtime.rs
@@ -40,8 +40,13 @@ pub const PODMAN_KUBE_RUNTIME_NAME: &str = "podman-kube";
 const CONFIG_VOLUME_SUFFIX: &str = ".config";
 const PODS_VOLUME_SUFFIX: &str = ".pods";
 
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, Default)]
 pub struct PodmanKubeRuntime {
+    workload_info_store: PodmanKubeWorkloadInfoStore,
+}
+
+#[derive(Debug, Clone, Default)]
+struct PodmanKubeWorkloadInfoStore {
     workload_info_cache: std::sync::Arc<Mutex<HashMap<String, PodmanKubeWorkloadId>>>,
 }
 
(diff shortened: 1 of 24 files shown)


agent/src/runtime_connectors/runtime_connector.rs
```diff
@@ -238,22 +252,21 @@ pub mod test {
...
-    #[derive(Debug)]
     pub enum RuntimeCall {
         GetReusableWorkloads(AgentName, Result<Vec<ReusableWorkloadState>, RuntimeError>),
         CreateWorkload(
             WorkloadNamed,
             Option<String>,
             Option<PathBuf>,
             HashMap<PathBuf, PathBuf>,
-            Result<(String, StubStateChecker), RuntimeError>,
+            Result<(String, StateCheckerHandle), RuntimeError>,
```
krucod3: Wouldn't it make sense to also change the String here to a `RuntimeWorkloadId`.

This would remove the mapping calls below.

agent/src/runtime_connectors/runtime_connector.rs
```diff
@@ -213,18 +237,8 @@ pub mod test {
     }
 
     #[async_trait]
-    impl StateChecker<String> for StubStateChecker {
-        fn start_checker(
```
krucod3, line 217: Funny that the `start_checker` was actually never user, just dragged around...
Great that you removed it now 👍

agent/src/runtime_connectors/unsupported/unsupported_runtime.rs
krucod3, line 1: I know this is not introduced here, but since we are touching it, I think it would make sense to move the `unsupported_runtime` and the `dummy_state_checker` to a new separate folder?

agent/src/runtime_connectors/podman_kube/podman_kube_runtime.rs
```diff
@@ -232,6 +323,7 @@ impl RuntimeConnector<PodmanKubeWorkloadId, GenericPollingStateChecker> for Podm
             manifest: workload_config.manifest,
             down_options: workload_config.down_options,
         };
+        let runtime_workload_id = RuntimeWorkloadId::from(workload_id.name.to_string());
```
krucod3: Just a couple of lines before this we are creating the `PodmanKubeWorkloadId` and discarding the object after we extract the new `RuntimeWorkloadId`.

To improve performance (e.g. for deletion), we should add the `PodmanKubeWorkloadId` to the cache of the new `PodmanKubeWorkloadInfoStore`.

agent/src/runtime_connectors/podman_kube/podman_kube_runtime.rs
```diff
@@ -240,88 +332,73 @@ impl RuntimeConnector<PodmanKubeWorkloadId, GenericPollingStateChecker> for Podm
...
+            .map_err(|err| RuntimeError::CollectLog(err.to_string()))?;
+        let workload_info = PodmanKubeWorkloadId {
+            name: instance_name,
+            pods: None,
+            manifest: String::new(),
+            down_options: Vec::new(),
+        };
         let podman_kube_log_fetcher =
-            super::podman_kube_log_fetcher::PodmanKubeLogFetcher::new(&workload_id, options);
+            super::podman_kube_log_fetcher::PodmanKubeLogFetcher::new(&workload_info, options);
```
krucod3: Can't we instead change the interface of the `PodmanKubeLogFetcher::new(...)`?

Creating a partially valid object just to satisfy an interface that apparently does not use the invalid data is highly dangerous.

krucod3: I just checked the `PodmanKubeLogFetcher` 🤣 

I think we should update the new to not take arguments and add a nice comment here 😬

agent/src/runtime_connectors/podman_kube/podman_kube_runtime.rs
```diff
@@ -240,88 +332,73 @@ impl RuntimeConnector<PodmanKubeWorkloadId, GenericPollingStateChecker> for Podm
...
 
-    async fn delete_workload(
-        &self,
-        workload_id: &PodmanKubeWorkloadId,
-    ) -> Result<(), RuntimeError> {
+    async fn delete_workload(&self, workload_id: &RuntimeWorkloadId) -> Result<(), RuntimeError> {
+        let instance_name = instance_name_from_runtime_id(workload_id)
+            .map_err(|err| RuntimeError::Delete(err.to_string()))?;
+        let workload_id = self
+            .workload_info_store
```
krucod3, line 394: As mentioned in the other comment, the perfomance here would be almost the same as before if we cache the `PodmanKubeWorkloadId` on create.

krucod3 (commented):
Reviewed, findings documented.

agent/src/runtime_connectors/runtime_connector.rs
christoph-hamm: fixed

agent/src/runtime_connectors/unsupported/unsupported_runtime.rs
christoph-hamm: fixed

agent/src/runtime_connectors/podman_kube/podman_kube_runtime.rs
christoph-hamm: fixed

agent/src/runtime_connectors/podman_kube/podman_kube_runtime.rs
christoph-hamm: it is no `pub struct PodmanKubeLogFetcher()` and no new method.

agent/src/runtime_connectors/podman_kube/podman_kube_runtime.rs
christoph-hamm: fixed

krucod3 (approved):
LGTM 👍

PR #763, opened 2026-07-02
Fix to long  container name

christoph-hamm:
# Fix container names for old versions of nerdctl

For older versions of nerdctl, the container name can not exceed 76 characters. This leads to only extream short workload and agents names being allowed. To fix this, the PR shortens the hash part of the container name. The hash used internally in the WorkloadIntanceId is not shortend.

# Definition of Done

### DIFF
--- a/agent/src/runtime_connectors/containerd/nerdctl_cli.rs
+++ b/agent/src/runtime_connectors/containerd/nerdctl_cli.rs
@@ -222,7 +222,7 @@ impl NerdctlCli {
         // We store workload name as a label (and use them from there).
         // Therefore we do insist on container names in particular format.
         //
-        // [impl->swdd~containerd-create-workload-sets-optionally-container-name~1]
+        // [impl->swdd~containerd-create-workload-sets-optionally-container-name~2]
         args.append(&mut vec![
             "--name".into(),
             instance_name.short_workload_name(),
@@ -644,7 +644,7 @@ mod tests {
     }
 
     // [utest->swdd~containerd-create-workload-creates-labels~1]
-    // [utest->swdd~containerd-create-workload-sets-optionally-container-name~1]
+    // [utest->swdd~containerd-create-workload-sets-optionally-container-name~2]
     // [utest->swdd~containerd-create-workload-mounts-fifo-files~1]
     // [utest->swdd~containerd-create-mounts-workload-files~1]
     #[tokio::test]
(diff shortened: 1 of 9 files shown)


inf17101:
I review this.

agent/src/runtime_connectors/containerd/nerdctl_cli.rs
```diff
@@ -201,14 +201,13 @@ impl NerdctlCli {
...
-        workload_name: &str,
-        agent: &str,
+        instance_name: &WorkloadInstanceNameSpec,
         control_interface_path: Option<PathBuf>,
         workload_file_path_mappings: HashMap<PathBuf, PathBuf>,
     ) -> Result<String, String> {
         log::debug!(
             "Creating the workload '{}' with image '{}'",
-            workload_name,
+            instance_name,
```
inf17101, line 210: I suggest that we print the workload name only or the short workload name version.

krucod3: Actually this is consistent with the old behavior. (see https://github.com/eclipse-ankaios/ankaios/pull/763/changes#diff-ed51790e59a276192c1d95dc207658c5a7f71432eacb59bcf77ee30f02214c91R179)
The variable name was misleading bevor the change.

ankaios_api/src/ank_base/workload_instance_name.rs
```diff
@@ -79,6 +79,18 @@ impl WorkloadInstanceNameSpec {
         ))
     }
 
+    pub fn short_workload_name(&self) -> String {
```
inf17101, line 83: I think a swdd mentioning the short workload name and its use case might be good. Then we can also link the tests below.

agent/src/runtime_connectors/containerd/nerdctl_cli.rs
christoph-hamm: I checked and the output is the same as before. I would also argue that the correct name is still the one with the full ID. That is the information used in the state and also used for finding running workspace using the name label.

I would also prefer to keep it as it is.

ankaios_api/src/ank_base/workload_instance_name.rs
christoph-hamm: I have added a requirements for the shortened name and also linked it. I also updated the requirements for the usage of the name in the podman and containerd runtime.

agent/src/runtime_connectors/containerd/nerdctl_cli.rs
inf17101: The variable name was changed from `workload_name` to `instance_name` and I then was confused because the change signaled to me that now the instance name is printed and not like before the workload name.

inf17101 (approved):
LGTM 👍

PR #766, opened 2026-07-21
Add systemd runtime support for managing existing services

pypingou:
Add systemd runtime support for managing existing services

Implements a new systemd runtime connector that enables Ankaios to
manage existing systemd services alongside container workloads.
Uses the new non-generic RuntimeConnector/RuntimeStateGetter traits
with RuntimeWorkloadId and StateCheckerHandle, matching the simplified
pluggable runtime architecture.

Features:
- Start, stop, and restart existing systemd services
- Monitor service state via systemctl show
- Stream logs via journalctl integration
- Support for three desired states: running, stopped, restarted
- Smart restart handling: skips restart if service was restarted recently
- Transient restart workloads: always trigger even if manifest unchanged
- Input validation on service names (safe characters and unit suffix)
- Full test coverage (43 unit tests including error paths)

Architecture:
- SystemdCli: Wrapper for systemctl commands using CliCommand pattern
- SystemdRuntime: Implements RuntimeConnector trait
- SystemdStateGetter: Implements RuntimeStateGetter trait
- SystemdLogFetcher: Streams logs via journalctl with follow/tail/since/until
- SystemdRuntimeConfig: Parses YAML configuration with serviceName and desiredState
- Registered via Box<dyn OwnableRuntime> array (no macro)

Design decisions:
- No unit file creation/deletion (only manages existing services)
- No reusability tracking (always reconciles to desired state)
- No control interface support (can't inject env vars into existing services)
- No file mappings (no container abstraction)
- Polling-based state monitoring (500ms interval)
- Uptime-based restart deduplication for agent self-restart scenarios
- Server detects restart workloads via YAML parsing (not string matching)
- Server treats restart workloads as transient (always sends, even if identical)

Restart handling rationale:
When restarting the ank-agent service itself, the agent reconnects after
restart and receives the same restart workload again from the server.
Two mechanisms work together:
1. Agent-side: Check service uptime, skip if recent (prevents duplicate restart)
2. Server-side: Always send restart workloads, even if identical (allows repeated restarts)

This enables multiple restart attempts while preventing spurious restarts
during agent reconnection.

Example usage:
```
  workloads:
    nginx:
      runtime: systemd
      agent: agent_A
      runtimeConfig: |
        serviceName: nginx.service
        desiredState: running
```

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/agent/src/runtime_connectors/systemd/mod.rs
+++ b/agent/src/runtime_connectors/systemd/mod.rs
@@ -1,0 +1,19 @@
+// Copyright (c) 2024 Elektrobit Automotive GmbH
+//
+// This program and the accompanying materials are made available under the
+// terms of the Apache License, Version 2.0 which is available at
+// https://www.apache.org/licenses/LICENSE-2.0.
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+// WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+// License for the specific language governing permissions and limitations
+// under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+pub(crate) mod systemd_log_fetcher;
+mod systemd_cli;
+mod systemd_runtime;
+mod systemd_runtime_config;
+pub use systemd_runtime::{SystemdRuntime, SYSTEMD_RUNTIME_NAME as NAME};
--- a/agent/src/runtime_connectors/systemd/mod.rs
+++ b/agent/src/runtime_connectors/systemd/mod.rs
@@ -1,0 +1,19 @@
+// Copyright (c) 2026 Red Hat LLC
+//
+// This program and the accompanying materials are made available under the
+// terms of the Apache License, Version 2.0 which is available at
+// https://www.apache.org/licenses/LICENSE-2.0.
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+// WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+// License for the specific language governing permissions and limitations
+// under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+pub(crate) mod systemd_log_fetcher;
+mod systemd_cli;
+mod systemd_runtime;
+mod systemd_runtime_config;
+pub use systemd_runtime::{SystemdRuntime, SYSTEMD_RUNTIME_NAME as NAME};
--- a/agent/src/runtime_connectors/systemd/systemd_cli.rs
+++ b/agent/src/runtime_connectors/systemd/systemd_cli.rs
@@ -1,0 +1,209 @@
+// Copyright (c) 2024 Elektrobit Automotive GmbH
+//
+// This program and the accompanying materials are made available under the
+// terms of the Apache License, Version 2.0 which is available at
+// https://www.apache.org/licenses/LICENSE-2.0.
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+// WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+// License for the specific language governing permissions and limitations
+// under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+#[cfg_attr(test, mockall_double::double)]
+use crate::runtime_connectors::cli_command::CliCommand;
+
+#[cfg(test)]
+use mockall::automock;
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct SystemdUnitState {
+    pub active_state: String,
+    pub sub_state: String,
+    pub exit_code: Option<i32>,
+}
+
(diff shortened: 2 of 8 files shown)


agent/src/runtime_connectors/systemd/mod.rs
```diff
@@ -0,0 +1,19 @@
+// Copyright (c) 2024 Elektrobit Automotive GmbH
```
krucod3: Please check the copyright headers.

agent/src/runtime_connectors/systemd/systemd_cli.rs
```diff
@@ -0,0 +1,209 @@
...
+    pub exit_code: Option<i32>,
+}
+
+pub struct SystemdCli {}
+
+#[cfg_attr(test, automock)]
+#[allow(dead_code)]
+impl SystemdCli {
+    const SYSTEMCTL_CMD: &str = "systemctl";
+    /// Start a systemd unit
```
krucod3: Please remove useless comments.

agent/src/runtime_connectors/systemd/systemd_cli.rs
```diff
@@ -0,0 +1,209 @@
...
+            let uptime_output = tokio::fs::read_to_string("/proc/uptime")
+                .await
+                .map_err(|e| format!("Failed to read /proc/uptime: {}", e))?;
+
+            // /proc/uptime gives system uptime in seconds (first field)
+            let uptime_str = uptime_output.split_whitespace().next()
+                .ok_or_else(|| "Failed to parse /proc/uptime".to_string())?;
+            let system_uptime_secs: f64 = uptime_str.parse()
+                .map_err(|_| "Invalid uptime format".to_string())?;
+            let system_uptime_us = (system_uptime_secs * 1_000_000.0) as u64;
```
krucod3: Could be extracted to a helper function for better readability, e.g., `get_system_uptime`.

The function could return in seconds directly and we could convert `enter_us` (not a very good name by the way) to seconds beforehand. This workflow would allow us to use u32 instead of u64 and will reduce the aritmetics.

agent/src/runtime_connectors/systemd/systemd_log_fetcher.rs
```diff
@@ -0,0 +1,336 @@
...
+            }
+        };
+
+        #[cfg(not(test))]
+        return Ok(Self { child: cmd });
+        #[cfg(test)]
+        Ok(Self {
+            child: cmd,
+            stdout: None,
+            stderr: None,
```
krucod3: If we either ways return None here, we can remove the fields from the struct and just return (None, None) on line 122 (https://github.com/eclipse-ankaios/ankaios/pull/766/changes#diff-73b53c395966efe945457b5e3ce4e4e8e7f6ba519d218bc7edbfd98c1c38e720R122) below.

agent/src/runtime_connectors/systemd/systemd_log_fetcher.rs
```diff
@@ -0,0 +1,336 @@
...
+        let cmd = match cmd {
+            Ok(cmd) => Some(cmd),
+            Err(err) => {
+                log::warn!("Cannot collect logs for '{}': '{}'", workload_id, err);
+                return Err(RuntimeError::CollectLog(format!(
+                    "Failed to spawn journalctl: {}",
+                    err
+                )));
+            }
+        };
```
krucod3: Long match instead of `map_err`:
```suggestion
        let child = Some(cmd.map_err(|err| {
            log::warn!("Cannot collect logs for '{}': '{}'", workload_id, err);
            RuntimeError::CollectLog(format!("Failed to spawn journalctl: {}", err))
        })?);
```

server/src/ankaios_server/server_state.rs
```diff
@@ -370,6 +381,19 @@ impl ServerState {
...
+                let transient = is_transient_restart_workload(workload);
+                if transient {
+                    log::debug!(
+                        "Removing transient restart workload '{}' from rendered_workloads",
+                        workload.instance_name.workload_name()
+                    );
+                }
+                !transient
+            });
+
```
krucod3: I think this was changed before we had the mutating hooks and the `effectiveState`. Which this change the `effectiveState` is not showing the restart workloads anymore.

krucod3 (commented):
I did a first review and will now think about how these one time workloads could be handled in a more consistent way.

krucod3:
The changes in the PR were done before the mutating hooks landed on main and the implementation breaks the effective state there, leaving transient restart workloads out of the effective state.

I added a commit to restore the originally intended behavior (or at least what I assume was the intended behavior 😉 ).
The transient restart workloads are now in the effective state, but they are not resent to agents that restart. This also has the effect that they are not sent to agents that connect after the the workload was added (which is also as intended). In this case the workloads are shown as `Pending(Initial)` forever. In the happy case the state is `AgentDisconnected` (also forever) if the target service was the ank-agent.service.

Somehow, the externally visible behavior looks patched.

Additionally, if not running the agent as root, one gets errors about "Interactive authentication". I also added a commit for using global flags with systemctl.

krucod3:
The more I look at the proposed runtime, the more I think it should actually not handle workloads, but tasks. I have to think a bit how we can bring this functionality in without breaking the concept that the runtime owns a workload.

PR #769, opened 2026-07-21
Fix couple of flaky tests

pypingou:
- Fix flaky ank_config tests (utest_ank_config_with_home_env_variable, utest_ank_config_without_home_env_variable) that raced on the HOME env var. Both tests read DEFAULT_ANK_CONFIG_FILE_PATHS, a Lazy static that initializes once per process — whichever test ran first froze the value, causing the other to fail. Extracted the path-building logic into a testable function so tests bypass the Lazy cache, and tests now save/restore HOME instead of leaving it removed.
- Fix flaky server_state test (utest_server_state_reapply_after_hook_mutation_no_update) that asserted on UPDATE_EFFECTIVE_STATE_CALLED, a process-global AtomicBool set by the update_effective_state test stub. Other tests calling ServerState::update() concurrently could set the flag between reset and assert. Removed the redundant check — result.unwrap().is_none() already validates no workload update was triggered.

Issues: #

# Definition of Done

### DIFF
--- a/ank/src/ank_config.rs
+++ b/ank/src/ank_config.rs
@@ -25,7 +25,6 @@ use grpc::security::read_pem_file;
 #[cfg(test)]
 use tests::read_pem_file;
 
-use once_cell::sync::Lazy;
 use serde::de::value::MapDeserializer;
 use serde::de::{Error, MapAccess, Visitor};
 use serde::{Deserialize, Deserializer};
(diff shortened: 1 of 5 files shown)


server/src/ankaios_server/server_state.rs
```diff
@@ -1922,10 +1921,6 @@ mod tests {
             result.unwrap().is_none(),
             "Reapplying a manifest after hook mutation shall not trigger a workload update"
         );
-        assert!(
-            !super::UPDATE_EFFECTIVE_STATE_CALLED.load(std::sync::atomic::Ordering::SeqCst),
-            "update_effective_state must not be called when no workload change is detected"
-        );
```
krucod3, line 1928: Admittedly, this is a double check here and we could drop the call.
This will fix the problem now, but somebody could fall into the same trap again is using the same global variable.

Let's switch to thread local variable and leave the double check in. It's not doing any harm after all.

We should switch to:
```Rust
#[cfg(test)]
thread_local! {
    static UPDATE_EFFECTIVE_STATE_CALLED: std::cell::Cell<bool> = const { std::cell::Cell::new(false) };
}
```

and use 

```Rust
super::UPDATE_EFFECTIVE_STATE_CALLED.with(|called| called.get())
```

and 

```Rust
super::UPDATE_EFFECTIVE_STATE_CALLED.with(|called| called.set(false))
```

in the tests.

krucod3 (commented):
Did not had time look at the ank_config.rs change yet, will do tomorrow.

server/src/ankaios_server/server_state.rs
pypingou: Adjusted :)

ank/src/ank_config.rs
krucod3: I looked a bit at the usage here and the lazy static is completely unnecessary.

When we are touching something here, I think it would be best to fix the underlaying problem.

Can you please remove the two lazy static variables:
* `DEFAULT_ANK_CONFIG_USER_FILE_PATH`
* `DEFAULT_ANK_CONFIG_FILE_PATHS`

and instead use your new functions `get_user_config_path()` and `get_config_file_paths()` directly?

Please just add the `output_warn!()` message to the else branch of the `get_config_file_paths` to maintain the same behavior.

pypingou: Adjusted

krucod3 (approved):
LGTM 👍

PR #784, opened 2026-08-27
Support unix domain sockets

### LINKED ISSUE #780: Support Unix Domain Sockets
## Description

Currently the server listens to an IP/port, by default 127.0.0.1:25551. If no mTLS is configured, every non-priviledged user on the same host can connect to that socket (e.g. via the ank CLI).
This insecure mode is intended for develoment and is the default for the installer and also for the packages. For development purposes a single node deployment might be sufficient and for that also Unix Domain Sockets (UDS) can be used as alternative to an IP connection. As UDS are file elements, those get permissions, users and groups. Thus normal Linux permissions schemes can be used to control access which is more simpler and provides better security than using IP without mTLS.

## Goals

- Enhance `ank-server`, `ank-agent` and `ank` to support UDS in addition
- The installer and the packages shall use UDS by default
- Restrict access to the Unix Domain Socket for Ankaios to root user and specific ankaios group, which is created by the installer and the packages. Thus every user who is member of the ankaios group can connect via UDS with `ank` to the `ank-server`.
- For the installer and package use case, `ank-agent` and `ank-server` shall also communicate via UDS. 

### Summary

**To be filled when the final solution is sketched.**

### Tasks

- [ ] Task 1
- [ ] Task 2
- [ ] ...

christoph-hamm:
Issues: #780

# Definition of Done

# ToDo

- [x] implement UDS in server
- [x] implement UDS in agent
- [x] implement UDS in CLI
- [x] adapt devcontainer
- [x] adapt installation script
- [x] adapt debian package
- [ ] check documentation
- [ ] get stests to work again in devcontainer

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -30,10 +30,10 @@ RUN echo 'alias ankaios-start="/workspaces/ankaios/tools/dev_scripts/ankaios-sta
 RUN echo 'alias ankaios-clean="/workspaces/ankaios/tools/dev_scripts/ankaios-clean"' | tee -a /home/${USERNAME}/.bashrc /home/${USERNAME}/.zshrc
 RUN echo 'export PATH=$PATH:/workspaces/ankaios/target/x86_64-unknown-linux-musl/debug' | tee -a /home/${USERNAME}/.bashrc /home/${USERNAME}/.zshrc
 
-# work in insecure mode in the dev container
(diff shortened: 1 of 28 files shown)


christoph-hamm:
The domains sockets have been added to the ank-server, ank-agent and ank. The devcontainer config, debian package and the install script have been updated as well. All three are working. Only for devcontainer the stests are not working at the moment.

I have not checked the documentation yet, if everything is still working. I think we need to mention the new ankaios group, and that you have to be part of that group in order to use the ankaios CLI. We can also remove the warning in the documentation.

